PR #21, opened 2012-10-20
Flexible python2

feth:
support creation of a virtualenv, using executables named python2, python2.7 or python2.6, whatever is available on the current system


feth:
pff, that's not what I meant. starting over

PR #23, opened 2012-10-20
blunt replacement of setup.py -> use d2to1

feth:
simply replaced setup.py by the d2to1 boilerplate

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,5 @@ bin/
 lib/
 local/
 include/
-venv_installed.stamp
-dev_env_installed.stamp
+.venv_installed.stamp
+.dev_env_installed.stamp
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ bin/
 lib/
 local/
 include/
+venv_installed.stamp
+dev_env_installed.stamp
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 PYTHON2 = `which python2 python2.7 python2.6 | head -1`
-DAYBED_EGG=lib/python2.7/site-packages/daybed.egg-link
+DAYBED_EGG=lib/$(PYTHON2)/site-packages/daybed.egg-link
 DEV_STAMP=dev_env_installed.stamp
 VENV_STAMP=venv_installed.stamp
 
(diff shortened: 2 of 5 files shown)


feth:
This now needs a correct setup.cfg file. Will do.

.gitignore
```diff
@@ -5,3 +5,5 @@ bin/
 lib/
 local/
 include/
+venv_installed.stamp
```
Natim: I think it could be better to create the file here: var/stamps/venv_installed.

Makefile
```diff
@@ -1,22 +1,36 @@
 PYTHON2 = `which python2 python2.7 python2.6 | head -1`
```
Natim, line 1: :+1:

Makefile
```diff
@@ -1,22 +1,36 @@
...
+DAYBED_EGG=lib/python2.7/site-packages/daybed.egg-link
```
Natim: :-1: If it is python2.6 the path will be wrong.

`find lib/ -name "daybed.egg-link"`

Natim:
Just two things but it looks good.

setup.cfg
```diff
@@ -1,27 +1,29 @@
-[nosetests]
-match = ^test
-nocapture = 1
-cover-package = daybed
-with-coverage = 1
-cover-erase = 1
+[metadata]
+name = daybed
+version=0.0
```
almet: should update to 0.1

setup.cfg
```diff
@@ -1,27 +1,29 @@
...
+description=daybed
+description-file=
+    README.rst
+    CHANGES.rst
+classifiers=
+    Programming Language :: Python
+    PrFramework :: Pylons
```
almet: "Pr" should be removed

setup.cfg
```diff
@@ -1,27 +1,29 @@
...
+    PrTopic :: Internet :: WWW/HTTP
+    PrTopic :: Internet :: WWW/HTTP :: WSGI :: Application
```
almet: ditto for these two

Natim:
Thank you feth

almet:
I second you on this, :+1:

almet:
No, really.

feth:
dimanche 2012-10-21 à 01:32:56 (CEST), Alexis Metaireau a écrit:

> I second you on this, :+1:

This would not have been possible were it not for alexis and natim.

almet:
What?

PR #27, opened 2012-10-23
Validate Spore again spore_definition + Add delete_data and validate it.

Natim:


daybed/designdocs.py
```diff
@@ -7,7 +7,7 @@
 """ Model definitions, by model name. """
 db_definition = ViewDefinition('name', 'definition', """function(doc) {
```
almet, line 8: if you don't emit the doc definition anymore, you'll need to change the name of the view and its description.

daybed/tests/spore_validation.rx
```diff
@@ -0,0 +1,89 @@
+{
```
almet, line 1: since this generation is handled by cornice, are we sure we want to add it here?

daybed/tests/test_functional.py
```diff
@@ -152,3 +173,10 @@ def test_data_validation(self):
         # of course, pushing weird data should tell what's wrong
         self.app.post_json('/data/todo', self.invalid_data,
                            headers=headers, status=400)
+        
+
+    def test_definition_deletion(self):
+        resp = self.create_definition()
+        self.app.delete('/definitions/todo?%s' % urlencode(resp.json))
```
almet, line 180: eeeew, isn't it weird to use directly the returned json here? I would expect create_definition to return me something else that only the id, what do you think?

daybed/validators.py
```diff
@@ -31,7 +31,7 @@ def schema_validator(request):
     model_name = request.matchdict['model_name']
 
     definition = request.db.get_definition(model_name)
-    schema = SchemaValidator(definition)
+    schema = SchemaValidator(definition['definition'])
```
almet, line 34: makes me thnk we should change definition to something else, that's weird to have a definition which contains something else.

dev-requirements.txt
```diff
@@ -1,6 +1,7 @@
-pyramid_debugtoolbar
-waitress
-lettuce
 WebTest
+lettuce
 nose
 nose-cov
+pyramid_debugtoolbar
+rxjson
```
almet, line 6: so, I don't think we need rx in the tests of daybed.

daybed/tests/spore_validation.rx
Natim, line 1: No that's true

daybed/tests/test_functional.py
Natim, line 180: Actually create_definition return the token.

almet, line 180: hmm, okay, I see. so you send something like ?token=XXX, that makes sense. A comment would help to understand :)

daybed/designdocs.py
Natim, line 8: Actually I emit the document which contains the definition. It does make sense in case you want to remove the definition or update it.

daybed/validators.py
Natim, line 34: It may be a better refactoring to rename the definition into schema the doc definition contain a schema.

daybed/tests/test_functional.py
Natim, line 180: https://github.com/spiral-project/daybed/blob/master/daybed/tests/test_functional.py#L89

PR #33, opened 2012-12-09
Geometry fields

leplatrem:


leplatrem:
Globally, I have doubts about this repeated `gps` parameter :(
I find it crucial to keep this option, but can't find a proper way to factorize its usage, from field to subnodes... Any help appreciated :)

PR #34, opened 2012-12-09
Add a view to list the fields.

almet:
Quite nothing, adding a list of fields at /fields.

Does that work for you?


leplatrem:
Ok for me.

If we think in terms of UI building, a couple of extra data will be necessary : description, parameters (regex, gps), etc. You could create an issue for later improvement :)

PR #49, opened 2013-02-22
Solve Issue 42

AntoineCezar:


leplatrem:
Curiously I was not notified of the PR :(

Globally I find this very neat ! :+1: 

... apart of course, hum hum, from the failing test (ImportError of `daybed.db` in tests), and the backend module name (I don't like the `_` in `couch_db` :v: )

daybed/backends/couch_db/backend.py
```diff
@@ -0,0 +1,47 @@
+import os
```
almet: I would name this couchdb and not couch_db

almet: would it be better to have everything in a couchdb.py file rather than having a module for it?

This means we should rename the backend to `CouchDBBackend` or something approaching

almet: After all, that makes sence to have everything in a couchdb package, like you did, since we have more than one file.

Still, I would like to rename this to couchdb and eventually expose a `CouchDBBackend` class rather than a generic `Backend`, for clarity.

tests.ini
```diff
@@ -1,5 +1,6 @@
 [app:main]
 use = egg:daybed
+daybed.backend=daybed.backends.couch_db.Backend
 couchdb_uri = http://localhost:5984/
```
almet: should we name couchdb_uri daybed.dbhost or something like that?

almet:
that's great, thanks!

appart some minor changes in the structure, I like it. +1

daybed/backends/couch_db/backend.py
AntoineCezar: Let's go with:

```
  daybed.backends.couchdb_backend.CouchDBBackend
```

tests.ini
AntoineCezar: I agree that _couchdb_uri_ should be name-spaced but i'll use _backend.dbhost_ to separate daybed settings from backend related ones.

almet: wfm.

daybed/backends/couch_db/backend.py
almet: "_backend" seem redundant here, isn't it?

AntoineCezar:
New version

daybed/tests/features/model_definition.py
```diff
@@ -28,7 +22,7 @@ def define_model_and_fields(step, modelaspect, model_name, fieldsaspect):
         'incomplete': '{"title": "Mushroom" %s}',
         'correct': '{"title": "Mushroom", "description": "Mushroom picking areas" %s}',
     }
-    
+
```
almet, line 25: thanks for removing these useless ones.

almet:
great.

Issue #53, opened 2013-05-30
Use JSON schema in definitions


leplatrem:
After discussion with @AntoineCezar and debates in issue #9 it seems relevant to use [JSON schema]([http://json-schema.org) for model definitions, instead of our own formalism.

A basic investigation needs to address the following questions:
- Should it replace or supplement the current definition formalism ?
- Does it support current and planned field types ? #51 #33 #29 
- Is it extensible and simple enough to add custom notions ? #11 
- Could it address the need of defining and validating relationships between models ? #32 
- Is it redundant with SPORE ?

More generally, should JSON schema validation come as a Cornice extension or as internal Daybed mechanics ?

AntoineCezar:
> Should it replace or supplement the current definition formalism ?

It could be plugable.

> Does it support current and planned field types ? #51 #33 #29

http://json-schema.org/latest/json-schema-validation.html#anchor108
for #33 we'll define the schema https://github.com/fge/sample-json-schemas/tree/master/geojson

> Is it extensible and simple enough to add custom notions ? #11
> Could it address the need of defining and validating relationships between models ? #32 

Play with this to get an idea: http://exavolt.github.io/onde/#?schema_url=schemas/jquery-package.json

AntoineCezar:
> Is it redundant with SPORE ?

SPORE Doesn't deal with schemas. Unless I'm missing something it's only about api discovery.

almet:
Thanks for bringing this to the table.

Using Json Schema seems indeed something to consider, using our own format wouldn't bring any good.
However, from what I see, JSONSchema still lacks some clear readability. By [looking at the documentation](http://json-schema.org/example2.html), I wasn't able to tell simply if it's possible or not to use regular expressions or not.

About the concerns you have:

### Should we replace or supersede what we have with JSONSchema?

I'm tempted to say we should use JSONSchema in parallel with the daybed formalism, but it means we need to write translation logic between the two, which I'm not sure would be reliable. If we validate JSONSchema as good enough to do what we are doing atm, maybe we should just switch there.

I've made a couple of queries to my search engine of choice and found that [someone already did something related](https://github.com/sbrauer/Audrey/blob/master/audrey/colanderutil.py#L3), so we may give it a go. (also, it means we lack the other direction, which is json schema to colander).

### Does it handle what we have or plan?

I'm unsure about that, but it seems pretty complete. But JSon Schema doesn't seem to be very easy to extend:

[...]

PR #55, opened 2013-05-31
Add required option to field definitions

leplatrem:
Allows to specify optional fields in definition.

I opend this PR, just to make sure you're ok with the `required` keyword at field level.

```
"fields": [{"name": "age", "type": "int", "required": False}]
```

If we switch to JSON schema #53, we'll move to global `required` attributes with list of field names, at definition level.

```
"fields": [{"name": "firstname", "type": "string"}, {"name": "age", "type": "int"}],
"required": ["name"],
```

The latter is slightly more complicated in client-side form generation...

### DIFF
--- a/daybed/schemas/base.py
+++ b/daybed/schemas/base.py
@@ -3,6 +3,7 @@ import datetime
 
 from colander import (
     deferred,
+    null,
     SchemaNode,
     Mapping,
     String,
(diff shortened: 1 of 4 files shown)


almet:
I prefer the former, but JSON Schema is probably a good idea, since they already define the format, so complying with their APIs is IMO the way to go, nope?

leplatrem:
Ok then.

We keep the former.

And think again about this once we implement #53.

You can merge if you like then.

PR #58, opened 2013-06-01
Add a POST method for definitions

AntoineCezar:
Lettuce tests fail but I don't know if it's still used. Is it?


leplatrem:
Great :) Activity !

Unfortunately many lettuce tests were not ported to unittest yet #31, so it's still relevant to keep them green ;)

Regarding the name generation, if you generate a uuid, it would be simpler to rely on couchdb, like it's done for data items : https://github.com/spiral-project/daybed/blob/master/daybed/views/data.py#L45-L49
However, it becomes very interesting if we generate the model id from the model title (a-la django slugify + loop for uniqueness)

AntoineCezar:
I've done it this way to allow other generation method. The originale use case wanted to generate human readable names.
Anyway there is still a problem : I've forgoten to return ne generated name.

AntoineCezar:
I've returned the name of the definition. I think returning the definition as stated in the issue wouldn't be consistant

leplatrem:
I'm no expert about REST conventions, but instead of returning the complete definition we could just add a redirect to the response like it's done on data post:
https://github.com/spiral-project/daybed/blob/master/daybed/views/data.py#L46-L48

AntoineCezar:
If we redirect. How to send the token ?
The client must actualy read the response, save the token then may follow the redirect. Sounds weird to me

AntoineCezar:
opened a ticket #61

AntoineCezar:
I should not rebase late on nigth after vine (Macvin du Jura) since I don't master rebase... Now I must find a way to clean the history

almet:
aha! Wine coding is always the best way to go.

Yeah, we probably better handle the tokens later on and keep things simple for now.

daybed/name_generators/random.py
```diff
@@ -0,0 +1,10 @@
+from uuid import uuid4
+
+
+class RandomNameGenerator(object):
```
almet, line 4: Oh, I was thinking about something which creates better names! Like that: https://github.com/seejohnrun/haste-server/blob/master/lib/key_generators/phonetic.js

Or something ala heroku, that's far better than the uuid4!

almet:
Otherwise sounds good to me!

leplatrem:
Ok, then we can return the full object and its URL as attribute instead of redirect, no ?

In the contrary, I would like to keep tokens on definitions until we find a proper way to implement ACLs :)

daybed/name_generators/random.py
AntoineCezar, line 4: I've made the key generator plugable on purpose. UUIDs works for me but other default generators are welcomed.
Should I rename the package "random-uuid" ?

AntoineCezar:
Created the dev branch

leplatrem:
Is that stuck until #61 is fixed ?

almet:
We've made a bunch of changes in the dev branch; I don' think that's too much hassle to port your changes there. I think we're ready for this.

almet:
Cherry-picked and updated with our latest changes. see e56cd08c7678d32f1472cf3323c17ed15fb1c594 and 85827ee32e6702d08045035d8fd3a95018e7b3f4

Issue #41, opened 2012-12-14
Implement PATCH on data items


almet:
PATCH allows us to only change some values inside an item, which is rather cool.

I want that :-) Less traffic, less headaches.

PR #62, opened 2013-06-08
Implement PATCH on data items (ref #41)

AntoineCezar:


leplatrem:
Since PATCH is supposed to implement partial updates, it would make sense to test just that.

Plus, should we be able to create items with PATCH ? If not, we should test that it's not possible ;)

leplatrem:
And if it is partial, I guess we can't use schema_validators at this place, but after record update...

almet:
Thanks! We've cherry-picked and merged it there 397f4dd14eaef15cb5dfe7c65de0bc988a91f0dd

PR #64, opened 2013-06-22
Refactoring

AntoineCezar:


almet:
I assume this is a work in progress? It seems to break the tests at the moment.

daybed/backends/couchdb/database.py
```diff
@@ -19,17 +27,36 @@ def get_definition(self, model_name):
...
         """Return the token associated with a definition.
 
         :param model_name: the name of the definition you want to retrieve
 
         """
-        return db_model_token(self.db)[model_name]
+        return db_model_token(self.__db)[model_name]
+
+    def delete_definition_data(self, model_name):
+        """Delete all data associatied of a model_name.
```
almet: "associated with"

daybed/backends/couchdb/database.py
```diff
@@ -19,17 +27,36 @@ def get_definition(self, model_name):
...
         """
-        results = db_definition(self.db)[model_name]
+        results = db_definition(self.__db)[model_name]
         for result in results:
             return result.value
 
+    def delete_definition(self, model_name):
+        """Delete the scheme definition from the model_name.
+
+        :param model_name: the name of the definition you want to retrieve
```
almet, line 37: "you want to delete"

daybed/backends/couchdb/database.py
```diff
@@ -10,8 +10,16 @@ class Database(object):
     """Object handling all the connections to the couchdb server."""
 
     def __init__(self, db):
-        self.db = db
-        self.save = db.save
+        self.__db = db
```
almet: why is this super private? I would have put a simple _db for it. Any reason?

daybed/backends/couchdb/database.py
```diff
@@ -37,21 +64,24 @@ def get_data(self, model_name):
         :param model_name: the name of the definition you want to retrieve
 
         """
-        return db_data(self.db)[model_name]
+        return db_data(self.__db)[model_name]
```
almet: It took me some time to understand what this was doing. `db_data` should probably be named `data_db` :-)

daybed/backends/couchdb/database.py
```diff
@@ -37,21 +64,24 @@ def get_data(self, model_name):
...
+
+    def _get_data_item(self, model_name, data_item_id):
+        key = [str(data_item_id), str(model_name)]
+        results = db_data_item(self.__db)[key]
```
almet: same here → `item_db`

almet: maybe is there a way to limit the number of items we want here? In case it returns more than one (for whatever reason)

daybed/backends/couchdb/database.py
```diff
@@ -37,21 +64,24 @@ def get_data(self, model_name):
...
 
     def get_data_item(self, model_name, data_item_id):
         """Get a data-item and checks it behaves to the requested model"""
-        key = [str(data_item_id), str(model_name)]
-        data_items = db_data_item(self.db)[key]
-        if len(data_items):
-            data_item = data_items.rows[0]
-            return data_item
-        return None
+        value = self._get_data_item(model_name, data_item_id)
```
almet, line 77: should we error out if we don't have any data associated, or returning None is enough?

daybed/views/data.py
```diff
@@ -3,7 +3,7 @@
 from cornice import Service
 from pyramid.httpexceptions import HTTPNotFound
 
-from daybed.validators import schema_validator
+from daybed.validators import data_validator
```
almet, line 6: thanks, that's better :)

daybed/backends/couchdb/database.py
AntoineCezar: The always wrong reason : anger. I saw "request.db.db" then I put on the stupid mode "WTF! I'm gonna kill this".

daybed/backends/couchdb/database.py
AntoineCezar, line 37: Found in views/definition.py

``` python
    results = request.db.get_data(model_name)
    for result in results:
        request.db.db.delete(result.value)

    result = request.db.get_definition(model_name)
    request.db.db.delete(result)
```

I've just avoided access to internal db

daybed/backends/couchdb/database.py
AntoineCezar: I don't understand the comment.

daybed/backends/couchdb/database.py
AntoineCezar, line 77: Using exceptions is in the pipe.

AntoineCezar:
The tests that break on my laptop are the two that always break:
- ERROR: test_date (daybed.tests.test_types.FieldTypeTests)
- ERROR: test_datetime (daybed.tests.test_types.FieldTypeTests)

I'll investigate this

daybed/backends/couchdb/database.py
AntoineCezar: :+1:

I tink having "view" in the name is a little better: data_view?

daybed/backends/couchdb/database.py
AntoineCezar: data_item_view?

daybed/backends/couchdb/database.py
almet, line 37: ah yeah I agree. My comment was about the docstring. I'm +1 to have this _db variable, but private is enough IMO - only one underscore

daybed/backends/couchdb/database.py
almet: there is a typo on the string I'm commenting: "associatied" please review my comments in the github PR and not via mail only since it can be misleading.

daybed/backends/couchdb/database.py
almet: is it a view? I though it was a database.

almet: Ah, a couchdb view. Wow, this is definitely misleading. data_db makes more sense to me, I think, so we avoid mixing terminologies.

almet:
Oh, I think this is tied to a fix we're waiting from colander, still not merged probably.

almet:
I think we've fixed most of the issues with https://github.com/spiral-project/daybed/commit/f74eb0e5df7dea9df8cd5a690cab5f106aabff4d + some other things. I'm closing this one, don't hesitate to propose other changes if you think we missed something.

PR #70, opened 2013-07-26
Store deserialized data

### LINKED ISSUE #67: Geometry values are serialized (and stored?) as string
See : 

http://daybed.lolnet.org/data/mushroomspots
http://daybed.lolnet.org/data/fefe

leplatrem:
I wonder how it took us so long without noticing.

We weren't storing the deserialized data, but the original posted data !
This preventing clean-up and defaulting values...

Specific case has to be handled manually for storing date/times into CouchDB. Best practice seems to serialize as ISO for `Date.parse()` to work natively. (source http://stackoverflow.com/a/4773320)

I propose this into master, since it is bug :) 

@natim, this may break daybed-pass since you are storing JSON as string. Luckyly, I had this check for a while in daybed-map https://github.com/leplatrem/daybed-map/blob/dev/models.js#L20-L22

This fixes #67

### DIFF
--- a/daybed/backends/couchdb/database.py
+++ b/daybed/backends/couchdb/database.py
@@ -1,3 +1,4 @@
+import datetime
 from .designdocs import (
     db_model_token,
     db_definition,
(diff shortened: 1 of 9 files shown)


leplatrem:
@Natim forget what I said. This won't affect daybed-pass, since you use basic string type.

AntoineCezar:
Is this a CouchDB specific matter?

leplatrem:
No it's not only related to CouchDB, at views level, we weren't using deserialized version of posted data.

But then, passing `date/time` objects to couchdb was failing. Therefore the specific piece of code at storage level (should be managed by couchdb driver itself IHMO)

PR #71, opened 2013-07-26
Store deserialized dev

leplatrem:
Same as #70, rebased.

### DIFF
--- a/daybed/backends/couchdb/database.py
+++ b/daybed/backends/couchdb/database.py
@@ -1,3 +1,4 @@
+import datetime
 from .designdocs import (
     db_model_token,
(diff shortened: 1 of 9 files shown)


almet:
I suggest we MERGE master, and not rebase it for this.

leplatrem:
As you want, but just be aware that it was not a piece of cake to rebase/merge this :)

almet:
The rationale behind this is that otherwise, we'll get two different branches that are not related to each other. And thus have a divergent history, which makes usually things complicated.

Let's try to avoid that.

almet:
See #80

PR #74, opened 2013-07-27
Add relations fields

leplatrem:
Most of the work here consisted in giving a reference to database from validators to field types checking. 
The rest is quite obvious.
# Behaviour

Can only reference an `id` of the linked model. Posting a record of the linked model directly is not supported.
# Code

Could not find any way to take a shortcut ! (app registry ? events ?)
# Storage

Wonder if we should serialize into somewhere more HATEOAS : 

Instead of : 

`abcdefgh`

store (and serve) this : 

```
{id: 'abcdefgh',
 model: 'simple',
 uri: '/models/simple/data/abcdefgh'}
```

Please comment :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,34 +1,35 @@
+VIRTUALENV=virtualenv
+VENV=.
+PYTHON=$(VENV)/bin/python
 DEV_STAMP=.dev_env_installed.stamp
-VENV_STAMP=.venv_installed.stamp
 INSTALL_STAMP=.install.stamp
 
 .IGNORE: clean
-.PHONY: functional_tests unit_tests tests
+.PHONY: all install virtualenv tests
 
 OBJECTS = bin/ lib/ local/ include/ man/ .coverage d2to1-0.2.7-py2.7.egg \
 	.coverage daybed.egg-info
 
 all: install
 install: $(INSTALL_STAMP)
-
-install-dev: $(DEV_STAMP)
-
-$(INSTALL_STAMP): $(VENV_STAMP)
-	bin/python setup.py develop
+$(INSTALL_STAMP): $(PYTHON)
+	$(PYTHON) setup.py develop
 	touch $(INSTALL_STAMP)
 
-$(DEV_STAMP): $(VENV_STAMP) dev-requirements.txt
-	bin/pip install -r dev-requirements.txt --use-mirrors
+install-dev: $(DEV_STAMP)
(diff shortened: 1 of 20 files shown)


almet:
I agree, this solution is kinda weird because we're passing the request to the SchemaValidator which in turn pass it to the other who needs it, but we only want to access the db instance.

I can see different ways to solve this:
- pass the db directly
- make the db a global object we can access from anywhere. In practice this is a bad idea but in theory having it implemented as a singleton can make things easier to manage and can stay thread-safe, and improve the readability of the code
- stay as you did it, in case we will want to have schema matching on other stuff that what we already checked in the request.

My personal perference goes to 2. since it means we need to touch less code, but I'm feeling that's kinda of hackish.

leplatrem:
I also prefer 2 !

I thought there was a way to obtain a global `application` instance, from there access `config`, then `registry`, then `db`...
But I didn't find any way that day...

almet:
Oh, good call. We actually can use `get_current_registry()`! http://docs.pylonsproject.org/projects/pyramid/en/1.0-branch/api/threadlocal.html#pyramid.threadlocal.get_current_registry

leplatrem:
Recommended to avoid usage outside tests:
http://docs.pylonsproject.org/projects/pyramid/en/1.0-branch/api/threadlocal.html#pyramid.threadlocal.get_current_registry

Singletons do not match pyramid design:
http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/introduction.html#no-singletons 

I thought I could make config object a daybed module attribute...
http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/introduction.html#automating-repetitive-configuration

But preferred the `global_registries` attribute of pyramid config module !
http://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html

leplatrem:
@AntoineCezar was not very enthusiast about this `from pyramid.config import global_registries` 

That's the only way to validate using decorators, and I like its brevity. Plus, it does not break any pyramid pattern AFAIK. 

Otherwise something similar to this has to be implemented: https://github.com/spiral-project/daybed/blob/5ba78/daybed/views/data_item.py#L52-L54

@ametaireau any objection ?

almet:
I don't have strong feelings either way: we should be careful about our use of the global registries, but I think this is a good usecase for it.

leplatrem:
done !

PR #76, opened 2013-07-27
Add multiple choices field type

leplatrem:


AntoineCezar:
Works for me

daybed/schemas/base.py
```diff
@@ -180,6 +184,39 @@ def validation(cls, **kwargs):
         return super(EnumField, cls).validation(**kwargs)
 
 
+class JSONList(List):
+    """A set of items in JSON-like format"""
```
almet: Why json _like_?

daybed/tests/test_types.py
```diff
@@ -74,6 +74,39 @@ def test_optional_field(self):
...
+
+    def test_choices(self):
+        schema = schemas.ChoicesField.definition()
+        definition = schema.deserialize(
+            {'name': 'tags',
+             'type': 'choices',
+             'choices': ['a', 'b', 'c']})
+
+        validator = schemas.ChoicesField.validation(**definition)
+        self.assertEquals(['a'], validator.deserialize('a'))
```
almet, line 97: I have share thoughts about the data we should be able to recognize. What's the rationale behind having non-json data?

almet:
This works for me. I would like to have some reasoning about the _like_ part. When would we want to have stuff looking like JSON but which isn't JSON?

leplatrem:
The only use-case I have in mind is with posting forms values.
For example, in most geometry editing applications, we fill a hidden textarea field with serialized JSON inside. When form is posted, the server receives the JSON data as string.

I also thought it would be an easy way to give multiple choices, as comma-separated : `values = "1,2,3"`

Hence, that "JSON-like" comment should be changed into something like "Pure JSON or string, as serialized JSON or comma-separated values"

almet:
+1

almet:
updated + merged.

PR #77, opened 2013-07-30
Makefile refactoring

AntoineCezar:


leplatrem:
Just one remark, not related to your changes : make serve from scratch does not run. make install has to be ran previously

AntoineCezar:
Good remark. It should works now

leplatrem:
Nothing against merging this on my side

Makefile
```diff
@@ -10,25 +12,24 @@ OBJECTS = bin/ lib/ local/ include/ man/ .coverage d2to1-0.2.7-py2.7.egg \
 
 all: install
 install: $(INSTALL_STAMP)
-
-install-dev: $(DEV_STAMP)
-
-$(INSTALL_STAMP): $(VENV_STAMP)
-	bin/python setup.py develop
+$(INSTALL_STAMP): virtualenv
```
almet, line 15: shouldn't this be $VIRTUALENV (in case it's over-written)?

almet:
+1 otherwise.

Makefile
AntoineCezar, line 15: Its not related to $VIRTUALENV. It could have been done this way:

```
$(INSTALL_STAMP): $(PYTHON)
```

PR #78, opened 2013-07-31
Tests refactoring

AntoineCezar:


daybed/backends/couchdb/database.py
```diff
@@ -58,17 +58,14 @@ def delete_data_item(self, model_id, data_item_id):
             self._db.delete(doc)
         return doc
 
-    def delete_data_items(self, model_id):
```
almet, line 61: I prefer to let the delete_data_items() method so we don't need to delete the model to delete the its data.

daybed/backends/memory/backend.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, config):
+        settings = config.registry.settings
+
+        self._db = {
+            'models': {},
+            'data': {}
+        }
+
+        # model id generator
+        generator = config.maybe_dotted(settings['daybed.id_generator'])
```
almet, line 20: maybe should we solve this in the includeme rather than here if we need to repeat it in each backend

daybed/backends/memory/database.py
```diff
@@ -0,0 +1,59 @@
+from copy import deepcopy
+
+
+class Database(object):
+    """Object handling all the connections to the couchdb server."""
```
almet: I guess that's a typo. this is not the couchdb server anymore.

daybed/tests/functional/test_todo_model.py
```diff
@@ -0,0 +1,41 @@
+from daybed.tests.support import BaseWebTest
```
almet, line 1: I'm not for or against having all the tests in the same file. What's the rationale behind the choice of splitting them into different files?

daybed/views/data.py
```diff
@@ -52,4 +52,4 @@ def post_data(request):
 @data.delete()
 def delete_data(request):
     model_id = request.matchdict['model_id']
-    request.db.delete_data_items(model_id)
+    request.db.delete_data_item(model_id)
```
almet: hmm, I want to delete the item_s_ from a model. Why changing this?

tests.ini
```diff
@@ -1,8 +1,6 @@
 [app:main]
 use = egg:daybed
-daybed.backend = daybed.backends.couchdb.CouchDBBackend
-backend.db_host = http://localhost:5984/
-backend.db_name = daybed-test
+daybed.backend = daybed.backends.memory.MemoryBackend
```
almet, line 3: If we change this, we need to make sure the CouchDB backend is tested; Otherwise we cannot be sure things aren't broken.

daybed/backends/couchdb/database.py
AntoineCezar, line 61: +1 I confused data/data_item

daybed/backends/memory/backend.py
AntoineCezar, line 20: I don't see any includeme, but you're right. There must be a better way.

daybed/tests/functional/test_todo_model.py
AntoineCezar, line 1: 500 lines starts to be fat too me. Maybe a matter of taste.

daybed/backends/memory/backend.py
almet, line 20: that's the **init** in views

leplatrem:
I would rather have splitted the tests by daybed module :  _test_base_fields_, _test_geom_fields_

:+1: for CouchDB testing ! Without it, I wouldn't have detected the serialization problems with datetime at storage level.

AntoineCezar:
Ok lets drop this

almet:
@AntoineCezar don't hesistate to open issues / discussions before implementing things, I know this might be upsetting for you to work on features we're discussing (and eventually rejecting) afterwards.

AntoineCezar:
I thought we agreed to decouple the backend.

leplatrem:
Maybe we can keep the MemoryBackend ? Well, appart from testing, I don't see how it can be useful without pickle/persistence...

Also, if there is a way to iterate over available backends and run all tests, that might become relevant. But honestly, we can't miss couchdb.

AntoineCezar:
MemoryBackend was only for tests. If the tests can't be untied from a real backend there is no point having this.

leplatrem:
Well, they can ! But we then must test all available backends :)

almet:
Yeah, for instance if we want to be sure to have really fast tests
most of the time we can use this backend while developping and add a
full test run as a pre-commit hook or run it by our CI.

If this makes the tests a lot faster, that may make sense to use them.
Also, if we add tests for each db backend, then I don't see any
problem to use this for the functional tests, for instance, but we
cannot replace the couchdb db in the tests for now because we miss
tests for it.

PR #83, opened 2013-08-07
Add memory backend

leplatrem:
Revamp of #78 with just MemoryBackend.

After discussion with @AntoineCezar, it makes sense to get rid of couchdb backend for testing.
However, we should have specific tests for each backend.

### DIFF
--- a/daybed/__init__.py
+++ b/daybed/__init__.py
@@ -10,6 +10,7 @@ import json
 
 from cornice import Service
 from pyramid.config import Configurator
+from pyramid.events import NewRequest
 from pyramid.renderers import JSONP
 from pyramid.authentication import (
     AuthTktAuthenticationPolicy, BasicAuthAuthenticationPolicy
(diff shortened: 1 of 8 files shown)


daybed/validators.py
```diff
@@ -35,11 +36,34 @@ def schema_validator(request):
 
 
 def validate_against_schema(request, schema, data):
+    """Validates data against specified schema, and attach
+    a deserialized and cleaned-up version of data to the request.
+    """
     try:
-        request.data_clean = schema.deserialize(data)
```
almet: better naming, thanks.

almet:
We need to implement tests for the CouchDB backend before we can merge this.

Also, can we have tests for the new MemoryBackend?

AntoineCezar:
I'll write them, if not assigned.

daybed/validators.py
leplatrem: was that a recommandation ? 'cuz `data_clean` attribute was not renamed in this commit...

almet: sorry if that was misleading, I'm just saying that using data_pure and data_clean seems better to me than the previous version

almet:
We've been adding the tests in the ACL branch :-)

almet:
We now have proper tests for the backends, feel freee to update this code to the new interfaces exposed and do another pull request :)

leplatrem:
Unfortunately backend tests were done in ACL branch :)

I suggest we wait for ACL to be merged in master. And then, just add a `BackendTestCase` to this same PR and merge it !

almet:
oh, right. +1 to this. @Natim do you have time to finish the work on ACL or should I take a stab at it?

Natim:
Now that it has been merged :) We can refactor to test both CouchDB Backend and memory backend.

leplatrem:
I'll take this one, before the end of next week :)

almet:
Really cool. When this will be done, please switch the tests to this backend by default so we avoid spending so much time during network calls. That should speedup the tests.

PR #85, opened 2013-09-09
ACL - Work in progress

Natim:
The branch is almost ready.
- [x] We already unit tested the backend to be able to implement other backends.
- [x]  We still need to make sure that the views tests passes.
- [x] We would like to implement the persona backend.
- [x] Add a token-based authentication
- [x] Make the tests use a different authentication than persona.
- [x] Rebase / Merge with master.


leplatrem:
Added checkboxes ;) github <3

Natim:
2 over 3 :D

almet:
I'm actually unsure about what kind of auth we should use for the API. I've been investigating Hawk (https://npmjs.org/package/hawk) but maybe that will be too complicated for the clients.

The other possibility would be to use a different and well known auth system such as a bearer token (as defined by OAuth 2.0) or a bearer token using Digest (but we would be subject to replay attacks).

Anyone  with other ideas? (I've successfully plugged persona auth + multiauth on my local repo so we're almost there).

leplatrem:
I'm pretty unexperienced with all these...

But I would say that Github does it well :)

The python client PyGithub was very easy to use, and the JS client geojson.io that stores gists on Github looks pretty simple too : 
https://github.com/mapbox/geojson.io/blob/gh-pages/src/core/user.js

Natim:
I like the idea of doing it simply with an AccessToken generated for each user.

Natim:
Today is the day of the big merge.
- [x] Fix tests
- [x] Merge Master
- [x] Merge dev
- [x] Verify that it works as before
- [x] Tag current master as daybed 0.1
- [x] Release new daybed as 0.9

leplatrem:
w00t ! Great ! 

Please make sure that the default policy is the same as the current one (read/write for anonymous)

Good luck !

almet:
Let's check it works _before_ doing the merge, huhu :-)

Since we're introducing incompatible API changes, maybe should we release Daybed 1.0? (as suggested by  [semver](http://semver.org/)).

Or maybe am I missing something about the versioning you're refering to?

Natim:
Ok guys,

I merge both master and dev here.
It was quite a work but we are almost there.

I have two tests that are failling but I don't know either why or if it is a real bug.
- For the LinearRingNode the code is here
- For the data missing in GeoJSON field I don't know either.

Can @leplatrem have a look and fix what is needed ?

All other tests seams to work.
I think a big review is needed.

When all test pass, I will read the documentation and verify that Installation and Examples are up-to-date.

The missing part, is anonymous access that we still have to handle.
For now, either Persona or BasicAuth works but not Anonymous access yet.

leplatrem:
First was due to https://github.com/mozilla-services/cornice/issues/150. I committed a fix.

Other problems appeared because changes brought in https://github.com/spiral-project/daybed/pull/70 were not ported into your new code. (for e.g. storing `request.data_clean` instead of `json.loads(request.body)`).

I agree to merge ACL without anonymous support in `dev` branch. I'll give you a hand to implement it before we merge into `master`.

Natim:
With request.data_clean I get `TypeError: <colander.null> is not JSON serializable`

Natim:
Ok I handled it.
We still have an error. @leplatrem do you know if it is a true bug or just the fact that the API changed from previous version and that we should modify the test.

I don't know if 'data' is part of the geojson format or if everything is good so far.

```
AssertionError: 'data' not found in 
{
    "features": [
        {
            "geometry": {
                "coordinates": [
                    [
                        [
                            0, 
                            0
                        ], 
                        [
                            0, 
                            1
                        ], 
                        [
                            1, 
                            1
                        ], 
                        [
                            0, 
                            0
                        ]
                    ]
                ], 
                "type": "Polygon"
            }, 
            "id": "0c4a92ac24704a65aee5ff79dece8b64", 
            "properties": {
                "mushroom": "Boletus"
            }, 
            "type": "Feature"
        }
    ], 
    "type": "FeatureCollection"
}
```

Natim:
Ok I get it, it seams that before, in geojson, we haven't the same result that in json and that we now have geojson in both.

Natim:
Oh dear I think we made it !

Natim:
Well @leplatrem I let your merge into dev then :)

leplatrem:
Merged manually to split formating and packaging stuff

Natim:
You forgot something, the Python3 support for daybed :)

leplatrem:
I opened a dedicated pull request for this. Was this related to ACL ?

Natim:
Yes it is part of the big merge between dev, master and ACL.
I think I am ok now, we can split properly if you think it is better.

leplatrem:
Well I didn't see it as a requirement, so I prefered to split... Sorry for the confusion :)
Smaller pull requests are easier to merge anyway :)

PR #95, opened 2014-01-02
Get rid of separate modules for list and detail views

leplatrem:
Never made sense to me.

Plus, it revealed inconsistencies in services naming and descriptions.


almet:
Agreed, this is far more easy to work with.

almet:
Although, I'm not sure how this will conflict with the ACL branch, so I suggest we merge it after we merge the ACL one (since @natim is doing this today).

Natim:
OMG

PR #99, opened 2014-01-04
Move JSON compatibility right along couchdb storage

leplatrem:
Since JSON serialization is specific to CouchDB backend, I moved it there.

If you don't agree, please argue :)


Natim:
I agree :+1:

Natim:
Already merged.

Issue #102, opened 2014-01-05
Allow anonymous usage


leplatrem:
- Make sure anonymous users are able to use Daybed
- Default policy for models should be the simplest one

Maybe we could have a settings to restore the "all open" behaviour of version 0.1 ?

Natim:
I made it the default behavior.

PR #104, opened 2014-01-05
102 anonymous

Natim:
Refs #102 
- Add ACL log,
- Add anonymous behavior,
- Fix ACL on model creation,
- Add JSON Forbidden page,
- Add documentation.


Natim:
We need to allow the user the choose the default_policy from the .ini file.

Natim:
@leplatrem can you review?

PR #123, opened 2014-01-10
Move id_generators to backends module

leplatrem:
The day we have several random generators and several kind of generators, we can have as many as many python modules as we want...


Natim:
Ok why not.

PR #124, opened 2014-01-10
Refactor tests and increase coverage

leplatrem:


daybed/tests/test_backend.py
```diff
@@ -2,7 +2,7 @@
 try:
     from unittest2 import TestCase
 except ImportError:
-    from unittest import TestCase
+    from unittest import TestCase  # flake8: noqa
```
Natim, line 4: I think you may upgrade your flake instead.

daybed/tests/test_schemas_geom.py
```diff
@@ -4,31 +4,61 @@
...
+        self.schema = schemas.PointField.definition()
+        definition = self.schema.deserialize(
             {'name': 'location',
              'type': 'point'})
+        self.validator = schemas.PointField.validation(**definition)
 
-        validator = schemas.PointField.validation(**definition)
-        self.assertEquals([0.4, 45.0], validator.deserialize('[0.4, 45.0]'))
-        self.assertEquals([0, 45], validator.deserialize('[0, 45]'))
+    def test_coordinates_as_float_or_integer(self):
```
almet: That's not blocking, but I generally find it more readable to have the test method named after a sentence. `test_coordinates_are_deserialized_as_floats_or_integers`.

But that's a matter of taste, I believe. (regarding some other tests, that may be more consistent)

almet:
r+ LGTM

almet:
As a side note, I wonder if coveralls is doing its job the right way. It says +0.77% (almost nothing).

It has been disabled on getpelican/pelican because of its lack of useful information on the PRs.

leplatrem:
I disabled coveralls on pull-requests.

PR #125, opened 2014-01-12
Use constants for default user and policies

leplatrem:


Natim:
Nice Thx :)

daybed/acl.py
```diff
@@ -10,6 +10,19 @@
 from daybed import logger
 
 
+USER_EVERYONE = 'system.Everyone'
+USER_AUTHENTICATED = 'system.Authenticated'
```
almet, line 14: These `system.*` are already provided by pyramid, why not directly use them? (they are the same as defining constants)

almet:
I don't think using a constant is useful here. That's a concept coming from pyramid and they in some sense define the usage of this term.

If we want we can use constants rather than strings, see http://docs.pylonsproject.org/projects/pyramid/en/latest/api/security.html?highlight=everyone#constants

almet:
I wouId suggest we backoff this commit and do it again :)

PR #127, opened 2014-01-12
Rename to records (fixes #112)

### LINKED ISSUE #112: Rename data to collection
`data` is a generic term. Since a `data` contains `record`, we could change it to `collection` instead.

What do you think?

leplatrem:


almet:
r=@ametaireau. thx

PR #130, opened 2014-01-13
Human readable policy

leplatrem:


daybed/acl.py
```diff
@@ -73,27 +91,57 @@ def permission_mask(permission):
...
-        'get_record': 0x0400,        # R
-        'put_record': 0x0B00,        # C+U+D
-        'patch_record': 0x0200,      # U
-        'delete_record': 0x0100,     # D
+        'post_model': PERMISSION_CREATE,
+        'get_model':  {'definition': {'read': True},
+                       'records':    {'read': True},
+                       'users':      {'read': True},
+                       'policy':     {'read': True}},
+        'put_model':  {'definition': PERMISSION_CRUD,
```
almet: put shouldn't include read.

daybed/acl.py
```diff
@@ -58,7 +76,7 @@ def principals_allowed_by_permission(self, context, permission):
         raise NotImplementedError()  # PRAGMA NOCOVER
 
 
-def permission_mask(permission):
+def permission_required(permission):
```
almet: I feel like the comment of this function is not up to date anymore.

almet: `get_required_permissions` may be a better name for the function.

almet: Also, I realize that this could be a const rather than a function. Since it's not _doing_ anything else than returning a dict item.

daybed/acl.py
```diff
@@ -73,27 +91,57 @@ def permission_mask(permission):
...
+        'get_record': {'records': {'read': True}},
+        'put_record': {'records': PERMISSION_CRUD},
+        'patch_record': {'records': {'update': True}},
+        'delete_record': {'records': {'delete': True}}
     }
     # XXX Add users / policy management.
     return mapping[permission]
 
 
+def permission_mask(permission):
```
almet: maybe should be renamed to `get_binary_mask`.

daybed/acl.py
```diff
@@ -73,27 +91,57 @@ def permission_mask(permission):
...
+        byte = 0
+        if perm.get('create'):
+            byte += 8
+        if perm.get('read'):
+            byte += 4
+        if perm.get('update'):
+            byte += 2
+        if perm.get('delete'):
+            byte += 1
+        return byte
```
almet, line 127: That is probably awfully slow to do that for every request?

almet:
I like the fact that the user can directly submit JSON for the policies, and that should definitely stay like this.

That being said, I'm concerned by the fact that it may be slow to do the json → hex translation for each request. What about storing the binary mask and provide a way to translate it to human readable JSON?

leplatrem:
Thank you very much for your feedback ! 

We can benchmark this, and if this is really a problem, we can indeed perform mask <--> json translations in both ways. 
The idea of storing JSON instead of mask was to be able to browse (and modify) database raw content with no pain :) Nothing more.

daybed/acl.py
leplatrem: can you explicit this ? I can modify a record without read it ? What is the use-case ?

Natim: it is not only about a use case, it is about what is doing the view.
Do you have to read the data to put new data in your view?

By the way a use case can be a script that put captor information every 15 minutes. Even if you grab access to the script and read the (username/apikey) you don't want to get access to the data.

leplatrem: Got it ! Thanks :)

daybed/acl.py
leplatrem, line 127: I ran 100 000 times the operation dict --> mask.
It took in average 3.51 msec for empty dicts and 3.88 msec for full permissions dicts.

We hence may consider code brevity, but not performance IMO :)

Natim, line 127: Can't we use this:

``` python
def singlemask(perm):
    RIGHTS = ['delete', 'update', 'read', 'create']
    perms = [p for p in perm if perm.get(p)]
    byte = 0
    for p in perms:
        byte += 2**RIGHTS.index(p)
    return byte
```

daybed/tests/test_backend.py
```diff
@@ -38,7 +38,7 @@ def setUp(self):
             "description": "One optional field",
             "fields": [{"name": "age", "type": "int", "required": False}]
         }
-        self.policy = {'role:admins': 0xFFFF}
+        self.policy = {'role:admins': {'definition': {'create':True}}}
```
Natim, line 41: Is it normal that it is not the same defined policy?

Natim:
For me it looks interesting but the curl commands is much more verbose, we should allow to defined with dict as well as with CRUD

Something like that : `{"definition": "CRUD", "records": "CUD", "users": "C", "policy":"C"}`

daybed/acl.py
leplatrem, line 127: Yes, I don't mind, if you both prefer !
Note that code is shorter but execution goes from 3 msec with empty dict to 9 msec for full dict (still acceptable of course)

daybed/tests/test_backend.py
leplatrem, line 41: Was shorter to write, and policy does not matter in backend tests

daybed/acl.py
Natim, line 127: Your solution looks more efficient and more readable/understandable so let's keep it.

leplatrem:
Ah, you mean to post/put stuff with curl ? Yes, I get it, I also like the string representation !

When validating input `{"admin": {"definition": {"create": True}}` could be as valid as `{"admin": {"definition": "C"}}`. And we would store the full json version anyway... If we figure out how to it with colander, it could make sense indeed.

Do you think it's a blocker for merging this PR ? I'd rather add a ticket as improvement...
Computing binary masks manually was also a PITA :)

Natim:
Yes please rebase and I will wait for a last @ametaireau review on this one :)

daybed/acl.py
```diff
@@ -41,53 +90,48 @@ def permits(self, context, principals, permission):
...
 
 
-def permission_mask(permission):
-    """Returns the permission mask associated with a permission, so that it's
-    possible to do boolean operations with them.
+def get_binary_mask(permission):
+    """Transforms the permission as ``dict`` into a binary mask, so that it's
+    possible to do boolean operations with it.
+
+    A binary mask has four blocs of 4 bits, i.e. one CRUD mask for each domain.
```
almet: blocks

almet:
There is just one typo, other than that I'm okay with this way of doing things. +1 as well for having a string representation :) r=@ametaireau

PR #131, opened 2014-01-13
Do not use constants for default user and policies

leplatrem:


leplatrem:
Oh my god, I just realized that what @ametaireau told me is that there were constants in pyramid :( `pyramid.security.Everyone` and `pyramid.security.Authenticated`...

almet:
yes, that was it :(

Issue #122, opened 2014-01-10
Why do we need a /personna page ?


leplatrem:
Since Daybed is an API, why do we need `/personna` with Mako etc. ?

Natim:
Well, the basic answer is to enable persona authentication in daybed.

Persona needs a server to validate authentication and daybed needs to bind the persona user with the daybed user.

Mako is not really related to persona but it is a lightweight template system that was first default in pyramid, it is why we choosed it in the first place.

One of the aim I have for daybed is to have it as a Javascript APP only backend for localstorage synchronization.
In that case the /persona url is really useful.

I hope this is enough to answer your question.

leplatrem:
I still don't get why do we need HTML :(

As for Javascript apps, why JSON end-points are not enough ?

leplatrem:
Relates to #103

leplatrem:
We should definitely get rid of this page. It's useless for JS authentication !

almet:
+1, we should remove it.

almet:
Closing this for now, I'll create a new and updated bug about the next steps for auth.

Issue #103, opened 2014-01-05
Document authentication in Javascript


leplatrem:
Example, snippet, ... it should be very easy to use Daybed with Javascript.
If not, then get rid of authentication by default.

Natim:
In JS there is no problem because we have persona.

Natim:
It is as simple as http://localhost:8000/persona :)

leplatrem:
It looks rather simple, indeed (https://developer.mozilla.org/en/Persona/Quick_Setup)

I'll confirm it's as simple as you say when upgrading backbone-daybed to `dev` branch :)

leplatrem:
We shall have a look at http://oauth.io too :)

leplatrem:
Since login and logout views check for a number of things, I cannot see how to perform **remote** authentication in Javascript. 
- audience parameter restricts remote authenticated to a limited set of websites.
- crsf is checked (remotely we don't have the session token)
- cors is not enabled (ref #137)

In pyramid_persona, this is achieved through JS server side templating :( 
https://github.com/madjar/pyramid_persona/blob/master/pyramid_persona/utils.py#L29
https://github.com/madjar/pyramid_persona/blob/master/pyramid_persona/templates/persona.js#L17

How could we allow any JS website or mobile app to authenticated on our storage-as-a-service API ? Can we serve the session token as JSON ? Should we disable audience checks like CORS with `'*'` ?

As a reference, here is the remote HTML page (running on localhost:9999) that I use to login on daybed (running on localhost:8000) 
https://friendpaste.com/6Zj4fEIqshHqDs5Ptpd892
Like explained in https://developer.mozilla.org/en/Persona/Quick_Setup

almet:
The audience can be set to "*" if that's how we intend to use it. I guess in JS apps, people would have to set their audience to the server we are running, nope?

leplatrem:
In JS app, we don't specify the audience, so that would work indeed !

What do you think of : 
- adding CORS to `/login` ?
- providing crsf token in `/login` response headers ?

This way, remote authentication in JS would be : 
- obtain the Persona `assertion` using the [official Persona JS lib](https://developer.mozilla.org/en/Persona/Quick_Setup) (the one that prompts in a popup)
- perform a `HEAD` on `/login` to obtain the crsf token
- perform a `POST` on `/login` using both  `assertion` and `crsf` (requires CORS)

@ametaireau, could you start by adding CORS on login/logout views please ?
[...]

PR #138, opened 2014-01-17
Get rid of Mako and persona page (ref #136, #122)

leplatrem:
rationale : 
- daybed is an API
- there should be no mention of authentication method outside application init
- the persona page was buggy #136
- the persona page is useless for remote authentication, like pretended for a _storage as a service_ #122

Instead we should allow login and logout pages to be opened to the world :) cf. #103


almet:
We need to check if Hawk can be an alternative for users, if persona isn't alright; I'm not sure persona won't work, actually, but this need some time from us to identify how we could integrate with this service. I think we should do that before removing any authentication page we have for now.

— Alexis

leplatrem:
How does a login html page make sense on a Web API ? It's not a rhetorical question, I am wondering for real :)

I find Persona very convenient in our case ! Once CORS will be enable on `/login`, and once there is a way to obtain csrf before posting credentials, the implementation of Persona in Daybed will be just fine !

Natim:
I guess as an example of JS persona implementation. As an easy backend to get started faster.
I used this template for persona authentication in my JS example apps.

leplatrem:
After a rather long discussion with @ametaireau yesterday. This page is relevant because : 
- When prompting credentials, Persona associates authentication to a specific website
- A client application would ask its users to visit this daybed page in order to authenticate (with possible redirect on success)
- Further calls on daybed API from client application will be authenticated using cookies set on daybed domain.

We can imagine removing this page, only if we confirm the following assomption : 

> When prompting for Persona credentials on the client application (using Persona JS lib), the assertion obtained 
> can  be posted on _daybed_ `/login` view to authenticated the user.
> This may depend on the audience value to '*' specified on daybed.

Once this confirmed, we can update #103 and rethink about this one :)

almet:
Since we last discussed that in january, things have evolved a bit.  Persona is not developed actively at Mozilla and its relevance for Daybed is not really clear.

One reasonable course of action seem to ditch persona out and implement a "normal" accounts system with a way to authenticate with hawk.

PR #145, opened 2014-02-20
Geojson serializer tests

leplatrem:
- [x] Fix with empty collections
- [x] Fix with JSONP
- [x] Add tests for GeoJSON field type (will fail with current code)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -13,6 +13,10 @@ deps =
     mock
 install_command = pip install --process-dependency-links --pre {opts} {packages}
 
+[testenv:py26]
+deps =
+    ordereddict
+
 [testenv:py33]
 deps =
     coverage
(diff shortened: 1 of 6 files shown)


leplatrem:
Tests fail mysteriously on python 2.6 and pypy :)

leplatrem:
Ready! @ametaireau @Natim pls merge after quick overlook!

tox.ini
```diff
@@ -13,6 +13,15 @@ deps =
...
 install_command = pip install --process-dependency-links --pre {opts} {packages}
 
+[testenv:py26]
+deps =
+    coverage
+    nose
+    webtest
+    unittest2
+    mock
+    ordereddict
```
almet: we probably want to put ordereddict in the dev-requirements as well

Natim: Or better add it in the REQUIREMENTS variable of setup.py only if python < 2.7
Actually using d2to1 it doesn't seams possible.

leplatrem:
Should be better now :) Thanks for spotting this, it was shitty

almet:
looks good to me :)

PR #146, opened 2014-02-24
Add ability to add title and descriptions to policies (fixes #121)

### LINKED ISSUE #121: Add a description string to policy
This would allow to describe a policy in human-readable way.

For example : 

```
 { 
       "description" : "Only admins and owners of records can modify existing records",
       "permissions": {
           "role:admins": 65535,
           "system.Authenticated": 32768,
           "system.Everyone": 17408
       }
}
```

leplatrem:


daybed/schemas/validators.py
```diff
@@ -47,7 +47,11 @@ def _crudSchema(self, domain):
     def deserialize(self, cstruct=null):
         if cstruct:
             self.children = []
-            for key in six.iterkeys(cstruct):
+            self.add(SchemaNode(String(), name='title', missing=drop))
+            self.add(SchemaNode(String(), name='description', missing=drop))
+            roles = [key for key in six.iterkeys(cstruct)
```
almet, line 52: too bad filter doesn't work with dicts :)

PR #151, opened 2014-05-26
Move JSON compatibility right along couchdb storage

Natim:


Natim:
#99 says it has been merged but I am not sure.
@leplatrem could you have a look and see if it matches what you wanted?

leplatrem:
Yes I guess that's ok.

I think that each backend should be in charge of formatting the input data to its own formalism.

However, since JSON serialization of `colander.null` or `datetime` might occur in several backends, we could have a couple of shared helpers methods.

Let's say we keep it this way so far, and refactor in case we add more JSON-based backends...

Natim:
So should we close or merge this PR?

elemoine:
How can this be considered mergeable by GitHub? It looks like the code has changed a lot since https://github.com/spiral-project/daybed/commit/4153e8aaf145ca119d2c9fb9e12327409eaffd94 – the put_data_item method does even exist anymore in the master branch as far as I can tell. Just saying… I may be wrong.

leplatrem:
Yes, forget about this one. It's already (in another form) in **master** : 
https://github.com/spiral-project/daybed/blob/master/daybed/schemas/validators.py#L73-L96

PR #155, opened 2014-05-26
Added setup instructions for OSX.

n1k0:
This one adds a new installation section to the docs, with specific setup for OSX. Another section for Linux should be added in a next PR.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -1,0 +1,81 @@
+Installing daybed
+=================
+
+Daybed has the following requirements:
+
+- A Python_ 2.6, 2.7, 3.x or PyPy_ installation
+- A CouchDB_ server instance running
+
+Daybed comes with a Makefile to simplify your life when developing. To install
+daybed in your current virtualenv and get started, just run::
+
+    $ make install
+
+Then, running the test suite is a good way to check that everything is going
+well, and is well installed. You can run them with::
+
+    $ make tests
+
+The test suite will run all the available tests for every supported Python
+environment. You can check the current build status
+`on Travis <https://travis-ci.org/spiral-project/daybed>`_.
+
+Installation on *n*x systems
+----------------------------
+
+Standard installation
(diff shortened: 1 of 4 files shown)


docs/source/install.rst
```diff
@@ -0,0 +1,58 @@
+Installing daybed
+=================
+
+Daybed has the following requirements:
+
+- A `Python <http://python.org/>`_ 2.6, 2.7, 3.3 or 3.4 installation
```
almet: Also, PyPy.

docs/source/install.rst
```diff
@@ -0,0 +1,58 @@
...
+    $ make serve
+
+Development installation
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you start hacking on daybed, a good practice is to ensure the tests keep
+passing for all supported Python environments::
+
+    $ brew install python python3 pypy couchdb
+    $ make install-dev
```
almet: install-dev is a dependency of "test", so you don't need this line.

docs/source/install.rst
```diff
@@ -0,0 +1,58 @@
...
+Then, running the test suite is a good way to check that everything is going
+well, and is well installed. You can run them with::
+
+    $ make tests
+
+The test suite will run all the available tests for every supported Python
+environment. You can check the current build status
+`on Travis <https://travis-ci.org/spiral-project/daybed>`_.
+
+Mac OSX
```
almet: It seems that the installation instructions are pretty similar on debian and macos, maybe we should just put the two next to each other here? (I'm always reluctant to have macos only installation docs)

almet:
Added a couple of comments. Sounds good overall.

docs/source/install.rst
Natim: I also removed py33 from the tox test matrice. I think we can keep only the last Python3 version for now.

almet: +1

n1k0:
Addressed comments.

PR #158, opened 2014-05-26
Modify the Makefile to use the current VENV or create one if not in a VENV

Natim:
This let the Makefile use the current VENV or create one locally if the user is not inside a VENV.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 VIRTUALENV=virtualenv
-VENV := $(shell echo $${VIRTUAL_ENV-.})
+VENV := $(shell echo $${VIRTUAL_ENV-.venv})
 PYTHON=$(VENV)/bin/python
 DEV_STAMP=$(VENV)/.dev_env_installed.stamp
 INSTALL_STAMP=$(VENV)/.install.stamp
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,31 +1,40 @@
-DEV_STAMP=.dev_env_installed.stamp
-INSTALL_STAMP=.install.stamp
+VIRTUALENV=virtualenv
+VENV := $(shell echo $${VIRTUAL_ENV-.})
```
Natim: This use the current VENV or default to .

Makefile
```diff
@@ -1,31 +1,40 @@
...
+PYTHON=$(VENV)/bin/python
+DEV_STAMP=$(VENV)/.dev_env_installed.stamp
```
Natim, line 3: We create the stamp for each VENV so that we can have multiple VENV and be sure that they are ready.

Makefile
almet: wouldn't it be better to default to .venv or something like that? so we don't clutter the current dir?

Natim: Yes definitly :+1:

Makefile
```diff
@@ -1,31 +1,39 @@
...
 install-dev: $(DEV_STAMP)
-$(DEV_STAMP):
-	pip install -r dev-requirements.txt
+$(DEV_STAMP): $(PYTHON)
+	$(VENV)/bin/pip install -r dev-requirements.txt
 	touch $(DEV_STAMP)
 
+virtualenv: $(PYTHON)
+$(PYTHON):
+	$(VIRTUALENV) $(VENV)
```
almet: why not put directly "virtualenv" here and use a variable?

Issue #150, opened 2014-05-26
Rethink authentication


almet:
Currently, daybed uses persona for authentication.

It seems that's not exactly what we want, since persona can only be used in a web context for now. Also, the project is not anymore actively maintained by mozilla, so this is not likely to change in the near future.

My proposal:
- Create an endpoint for users to create accounts;
- Let them authenticate with different authentication schemes: Digest and Hawk authentication.

In the future, we could consider generating multiple hawk tokens for the users, so they can delegate access to their accounts to 3rd parties. Also, Hawk has a concept of "bewits" for this specific use case, that we may consider.

But I would prefer to keep this as a post 1.0 feature, so we can focus on what's important for now: having authentication that just works.

almet:
I'm wondering if that could make sense to actually integrate with Velruse, which proposes a way to integrate with 3rd parties. See http://velruse.readthedocs.org/en/latest/index.html

Thoughts?

almet:
We've been talking with @natim about authentication, and it seems that we need to solve different problems.

## How one logs into daybed?

(That's what we were solving before with the persona integration.)

We want to be able to log from different identity providers (github, facebook, google, twitter, SAML etc.). We don't want daybed to deal with password and email validation, for instance.

## How do we delegate daybed data access to 3rd party HTML apps?

In other words: how can an application get hawk credentials?

Here is a tentative flow:
- An user is directed to the daybed website and prompted to authorize access to a 3rd party app (a list of models to access is prompted);
- The user clicks on "YES"
- Hawk creds are sent back to the app (see implementation hereafter).

### Web Socket implementation
- The client creates a Web Socket connection with the server;
- It requests an UUID from the server and stores it;
- When the user is directed to the authorize page, the page is given this UUID;
- When the user clicks on YES, the server is notified, the server generates hawk credentials, finds back the right websocket connection with the UUID and sends it back the hawk credentials via the web socket.

### Simple Push

Quite similar, but not exactly (see https://gist.github.com/matzew/cbda360d72eaaef75971):
- You create the websocket
- You request an URL
- You give this URL to Daybed
- Daybed stores the hawk credentials somewhere and call the URL with a version number
- It wakes up the client using the websocket
- Client asks credentials using its Simple Push URL + Version number.

### Another (maybe simpler) way: SSE

Not sure how exactly yet, see https://developer.mozilla.org/en-US/docs/Server-sent_events/Using_server-sent_events

## How do we delegate daybed data access to 3rd party server apps?

[...]

PR #160, opened 2014-05-26
(WIP) Add docker deployment

leplatrem:
- [x] Merge with n1k0 install docs
- [x] Use environment variables
- [x] Publish 1.0.dev image on docker.io


elemoine:
Nice! Out of curiosity, what's the exact goal of using Docker? Is it for daybed developers to quickly have a working dev environment? Or for people who want to deploy daybed for an actual use?

Btw here's a interesting article that suggests using an init process in Docker containers: http://phusion.github.io/baseimage-docker/, which sounds wise to me.

Natim:
@elemoine It is more for really fast user deployement for the next release we will release the docker image pre configured as well as a Debian package to make it really fast and easy to deploy.

conf/production.ini
```diff
@@ -0,0 +1,53 @@
+[app:main]
+use = egg:daybed
+# Backend configuration
+daybed.backend = daybed.backends.couchdb.CouchDBBackend
+backend.db_host = http://couchdb:5984/
+backend.db_name = daybed
+persona.secret = THIS IS SUPER SECRET.
+persona.audiences = http://localhost:8000 http://0.0.0.0:8000 http://127.0.0.1:8000
```
Natim: The user needs to change that for production.

leplatrem: If we remove the Persona stuff, nothing to change in production !

leplatrem:
Thanks Eric for getting in touch :)

I can see two main use-cases of Docker for Daybed : 
- Frontend developers who don't want to bother with python and stuff. Just get a local Daybed instance running ; 
- Easy deployment of Daybed instances in production, with the idea that every Mozilla service can easily be deployed at home :)

Regarding init in Docker, we've had this discussion for a while at Makina. The main disavantage is the lack of cron. But we can survive this by logging to stdout and delegate log handling to docker host daemon...

conf/production.ini
Natim: We will see but in the documentation there is a way to override the conf so it works for me.

Natim:
Why is it a WIP? What are you expecting to add?

conf/production.ini
almet: @leplatrem it seems that we finally will not remote persona, see #150 for more background on the why.

almet: s/remote/remove

leplatrem: Ok, we don't remove Personna, then I'll use environment variables to set this up.

leplatrem:
Not WIP anymore :)
Ready to review/merge

**Notes:**
UWSGI supports environment variables substitution in ini files for the `[uwsgi]` section (`$(VAR)` syntax) . But Paste config parser does not expand them at all in the other sections, so I did it in python with `os.path.expandvars` (`${VAR}` syntax).

daybed/__init__.py
```diff
@@ -53,9 +54,17 @@ def get_user(request):
     return request.db.get_user(userid)
 
 
+def settings_expandvars(settings):
+    """Expands all environment variables in a settings dictionary.
+    """
+    return dict((key, os.path.expandvars(value))
+                for key, value in settings.iteritems())
```
Natim: Please use `six.items` instead of `iteritems` since it break Py3 compatibility

leplatrem:
Done :)

PR #161, opened 2014-05-29
List field type (fixes #54)

### LINKED ISSUE #54: Add list field type
List of integer, string, etc. or list of "sub-model" for inline records

leplatrem:
- [x] Basic field definition 
- [x] Implement and tests validators

### DIFF
--- a/daybed/schemas/list.py
+++ b/daybed/schemas/list.py
@@ -1,5 +1,6 @@
 from pyramid.i18n import TranslationString as _
-from colander import (Sequence, SchemaNode, Length)
+from colander import (Sequence, SchemaNode, Length, String, drop, OneOf,
+                      Invalid)
 
 from .base import registry, TypeField, JSONList
 
@@ -12,6 +13,19 @@ class ListField(TypeField):
     @classmethod
     def definition(cls, **kwargs):
         schema = super(ListField, cls).definition(**kwargs)
+        schema.add(SchemaNode(String(),
+                              name='subtype',
+                              validator=OneOf(registry.names),
+                              missing=drop))
         schema.add(SchemaNode(Sequence(), TypeField.definition(),
-                              name='subtype', validator=Length(min=1)))
+                              name='subfields',
+                              validator=Length(min=1),
+                              missing=drop))
+
+        def validator(node, value):
+            if 'subtype' not in value and 'subfields' not in value:
+                msg = u"No 'subtype' nor 'subfields'"
+                raise Invalid(node, msg)
+        schema.validator = validator
+
         return schema
--- a/daybed/schemas/list.py
+++ b/daybed/schemas/list.py
@@ -1,0 +1,17 @@
+from pyramid.i18n import TranslationString as _
+from colander import (Sequence, SchemaNode, Length)
+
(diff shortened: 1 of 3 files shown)


daybed/schemas/list.py
```diff
@@ -0,0 +1,31 @@
...
+                              name='subtype',
+                              validator=OneOf(registry.names),
+                              missing=drop))
+        schema.add(SchemaNode(Sequence(), TypeField.definition(),
+                              name='subfields',
+                              validator=Length(min=1),
+                              missing=drop))
+
+        def validator(node, value):
+            if 'subtype' not in value and 'subfields' not in value:
```
almet: why not naming them "fields" and "fields_type"?

almet: (or list_fields)

leplatrem: I'm still not sure about the naming indeed. I shall look at forms libs and see what comes out.

The idea is to have 3 kinds : 
- list of basic types (int, string, ...) (`subtype` in the current code)
- list of objects with some fields (`subfields` in the current code)
- list of sub-model

It could be : 
- `itemtype` : expect an existing field type (`int`, `string`, ...) or `model` or `object`
- if `object`, then expect `itemfields`
- if `model`, then expect existing model name `itemmodel`

leplatrem: Or even better : 
- Add new field types `object` and `model`
- Define a list just as a sequence of an existing field type

An `object` field is defined by an optional list of `fields` (if no field specified, no specific validation appart from being some keys/values)

A `model` field is defined with an existing model name. Validation is made against the definition of this model.

leplatrem:
Ready to be reviewed.

Can be merged after PR #162

leplatrem:
Rebased on master for smaller diff

leplatrem:
Please review and merge

daybed/schemas/list.py
```diff
@@ -0,0 +1,54 @@
...
+class ParametersMatchItemType(object):
+    def __call__(self, node, value):
+        if 'parameters' in value and 'itemtype' not in value:
+            msg = u'Missing itemtype since parameters is provided'
+            raise Invalid(node, msg)
+        if 'itemtype' in value and 'parameters' in value:
+            itemtype = value['itemtype']
+            parameters = value['parameters']
+            schema = registry.definition(itemtype)
+            parameters.update(name='fake', type=itemtype)
```
almet, line 17: why do we need to use name=fake here?

leplatrem, line 17: Because we do not manipulate a real field here.

We just want to check that the `parameters` match the ones expected by this field type.

For example : 

```
{
     'name': 'datelist',
     'type': 'list',
     'itemtype': 'date',
     'parameters': {
          'autonow': True
     } 
}
```

Here we want to check that `autonow` is a valid parameter, and `True` a valid value for it. So we instantiate a fake field : 

```
{
     'name': 'fake',
     'type': 'date',
     'autonow': True
} 
```

And check that it is a valid field definition :)

almet:
Merged! (added https://github.com/spiral-project/daybed/commit/a0317cd43bc54bdf19cba53538f2e35a81f98a2d )

PR #170, opened 2014-06-16
Increase code coverage

leplatrem:
- Added some missing exceptions catchs
- Added some corner cases in object field tests


almet:
Thanks!

PR #171, opened 2014-06-17
Rework permissions documentation

leplatrem:
I spent some time to improve the permissions docs. Basically, I took what @Natim had written and : 
- Removed the acronym "ACL" (it may be rather RBAC?)
- Made sure we use a consistent vocabulary
- Removed internal notions as much as possible
- Removed JSON notation when it wasn't about API

This changes highlight the following points : 
1. Permissions are hard to explain. Thus even harder to understand.
2. Permissions on `roles` and `policy` are never checked individually (thus superfluous)
3. It might be clearer if default `role:admins` was renamed to `role:owners`
4. It would be wise to substitute `system.Authenticated` by `role:authenticated` and `system.Everyone` by `role:everyone`


docs/source/permissions.rst
```diff
@@ -0,0 +1,211 @@
...
+For example, the following one allows manipulation
+of records only :
+
+==========  ======  ====  ======  ======  ============
+            Create  Read  Update  Delete  *(Shortcut)*
+==========  ======  ====  ======  ======  ============
+definition          True                  -R--
+records     True    True  True    True    CRUD
+policy              True                  -R--
+roles               True                  -R--
```
almet, line 42: Thanks for taking the time to update this. However, it seems that we are moving away from this approach now. e.g. when the refactoring of the ACLs will be done, we'll not have any roles anymore.

leplatrem, line 42: I don't care about the time I spent. It was necessary for me to understand the current situation.

I would definetly appreciate if you start a branch and start rewriting the documentation with the new system you have in mind.

This way, everybody would understand the direction you want to take, and the changes it brings in term of concepts and API...

almet, line 42: That's indeed a good proposal, docs driven development!
Plan is to take time next week to work on daybed, that will be the first thing then :)

PR #173, opened 2014-06-19
[WIP] ElasticSearch implementation

leplatrem:
- [x] Create index mapping on model creation
- [x] Index on record creation
- [x] Search view, proxy on E/S
- [x] Search with request body
- [x] Search with URL parameters
- [x] Reindex on record update
- [x] Index individual records on model records PUT
- [x] Test model definition update
- [x] Test model definition mapping (shapes, etc.)
- [x] Test model record mapping (object, etc.)
- [x] Test spatial query
- [x] Return specific response code when _elasticsearch_ not available/reachable


daybed/__init__.py
```diff
@@ -141,6 +142,57 @@ def add_db_to_request(event):
         event.request.db = config.registry.backend.db()
     config.add_subscriber(add_db_to_request, NewRequest)
 
+    # index initialisation
```
almet: should we split this out in a separate function?

daybed/views/models.py
```diff
@@ -42,6 +43,9 @@ def post_models(request):
         roles=request.validated['roles'],
         policy_id=request.validated['policy'])
 
+    event = ModelCreated(model_id, request)
+    request.registry.notify(event)
```
almet: what about having a `request.trigger_event` (or`notify`) utility function, which would have an `event_type` parameter?

We could then do something like `request.trigger_event("model_created", model_id, request)`.

(I find the current form to be a bit weird (using request.registry sound borked to me)

daybed/__init__.py
leplatrem: yes of course !

daybed/views/models.py
leplatrem: To be honest, it's the first time I use Pyramid events :) Yes, why not using Javascript-like events !

leplatrem: Note: http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/events.html#creating-your-own-events

(helper would still be relevant)

PR #185, opened 2014-07-29
Add PUT and PATCH on models/{modelname}/acls endpoint.

Natim:


daybed/schemas/validators.py
```diff
@@ -127,3 +130,38 @@ def model_validator(request):
...
+    for token, acls in six.iteritems(body):
+        error = False
+        perms = set([perm.lstrip('-').lstrip('+').lower() for perm in acls])
+        if "all" in perms:
+            perms = set(PERMISSIONS_SET)
+        if not perms.issubset(PERMISSIONS_SET):
+            request.errors.add('body', token, 'Invalid permissions: %s' %
+                               ', '.join((perms - PERMISSIONS_SET)))
+            error = True
+        if token != "Authenticated" and token != "Everyone":
```
almet: `token not in ("Authenticated", "Everyone")`

daybed/backends/memory/__init__.py
```diff
@@ -44,11 +44,15 @@ def __get_raw_records(self, model_id):
...
+    def get_records(self, model_id, with_authors=False):
         records = []
         for item in self.__get_raw_records(model_id):
             item['record']['id'] = item['_id']
-            records.append(deepcopy(item['record']))
+            if with_authors:
+                records.append({"authors": deepcopy(item['authors']),
+                                "record": deepcopy(item['record'])})
+            else:
+                records.append(deepcopy(item['record']))
```
almet: can we factorize this with the previous line?

daybed/backends/redis/__init__.py
```diff
@@ -56,10 +56,14 @@ def __get_raw_records(self, model_id):
...
-    def get_records(self, model_id):
+    def get_records(self, model_id, with_authors=False):
         records = []
         for item in self.__get_raw_records(model_id):
-            records.append(item['record'])
+            if with_authors:
+                records.append({"authors": item["authors"],
+                                "record": item["record"]})
+            else:
+                records.append(item["record"])
```
almet: ditto

daybed/backends/memory/__init__.py
almet: Actually, I find it very weird to have two different output format here, depending on the parameters passed to the function. We should return "record" in both cases.

docs/source/permissions.rst
```diff
@@ -1,211 +1,57 @@
...
+- **update_definition**: You can update the model definition
+- **update_acls**: You can change the ACLs on the models
+- **delete_model**: You can delete a model
+- **create_record**: You can add an entry to the model
+- **read_all_records**: You can read all model's records
+- **update_all_records**: You can update all model's records
+- **delete_all_records**: You can delete any model's records
+- **read_my_record**: You can read records on which you are an author
+- **update_my_record**: You can update and change records on which you are an author
+- **delete_my_record**: You can delete records on which you are an author
```
leplatrem: Wouldn't it make more sense to have `own` instead of `my` and `records` plural ?
(**read_own_records**)

Natim: Yes ok

daybed/__init__.py
```diff
@@ -84,7 +81,11 @@ def main(global_config, **settings):
     )
     config.set_authentication_policy(authn_policy)
     config.set_authorization_policy(authz_policy)
-    config.add_request_method(unauthenticated_userid, 'token', reify=True)
+
+    def get_token(request):
+        return request.unauthenticated_userid
```
almet: I'm not sure what this does. Maybe we should do request.authenticated_userid?

daybed/__init__.py
```diff
@@ -98,6 +99,12 @@ def add_db_to_request(event):
 
     config.add_subscriber(add_db_to_request, NewRequest)
 
+    def add_default_accept(event):
+        if "Accept" not in event.request.headers:
+            event.request.headers["Accept"] = "application/json"
```
almet, line 106: If this is to fix the geojson issue we are having, then we should add a comment about that :)

daybed/__init__.py
```diff
@@ -98,6 +99,12 @@ def add_db_to_request(event):
...
+
+    config.add_subscriber(add_default_accept, NewRequest)
```
almet, line 108: does this actually fixes the problem?

daybed/backends/couchdb/__init__.py
```diff
@@ -69,13 +69,21 @@ def get_model_definition(self, model_id):
...
         records = []
-        for item in self.__get_raw_records(model_id):
+        for item in raw_records:
             item.value['record']['id'] = item.value['_id'].split('-')[1]
-            records.append(item.value['record'])
+            if with_authors:
+                records.append({"authors": item.value['authors'],
+                                "record": item.value['record']})
+            else:
+                records.append(item.value['record'])
```
almet: my comment still applies: we should not have two different outputs depending on the parameters.

daybed/__init__.py
Natim, line 108: At least there is a test for it.

daybed/backends/couchdb/__init__.py
Natim: Maybe we can great `get_records_with_authors` instead?

almet: I wouldn't be chocked to have this function return a dict which could not contain the "authors" key. If you feel more comfortable with a "get_records_with_authors" function, I wouldn't be opposed to it ;)

Natim: It is just that get_records returns:

```
[{"age": 42, "id": "123456"}, {"age": 25, "id": "45789"}]
```

And I don't want to make it return:

```
[{record: {"age": 42, "id": "123456"}}, {record: {"age": 25, "id": "45789"}}]
```

daybed/schemas/validators.py
```diff
@@ -127,3 +122,38 @@ def model_validator(request):
         for record in records:
             validate_against_schema(request, definition_schema, record)
             request.validated['records'].append(record)
+
+
+def acls_validator(request):
+    """Verify that the acls defined are ok."""
```
almet: Hmm, this explanation text is a bit fuzzy to me. What does "are okay" means?

I would see something like the following: "Verify the ACLs defined in the request body are considered valid".

daybed/tests/test_acl.py
```diff
@@ -77,6 +77,21 @@ def test_acl_permits(self):
         self.assertTrue(permits(context, ['Alexis', Authenticated],
                                 'get_definition'))
 
+    def test_dict_set2list(self):
```
almet, line 80: maybe these functions finally belong to util.py…

Natim, line 80: I put them here because they are useful only for the acl_matrix in our case.

daybed/tests/test_views.py
```diff
@@ -91,6 +97,34 @@ def test_forbidden(self):
...
 
+    def test_unauthorized_token(self):
+        self.app.put_json('/models/%s' % self.model_id,
+                          {'definition': self.valid_definition},
+                          headers=self.headers)
+
+        auth = base64.b64encode(
+            u'foo:bar'.encode('ascii')).strip().decode('ascii')
+
+        resp = self.app.get('/models/%s' % self.model_id,
```
almet, line 108: Hmm, that remembers me we need to add a config telling if we want to accept basic auth or not. Created https://github.com/spiral-project/daybed/issues/186 about that.

daybed/tests/test_views.py
```diff
@@ -91,6 +97,34 @@ def test_forbidden(self):
...
+                            },
+                            status=401)
+        self.assertIn('401', resp)
+
+    def test_forbidden(self):
+        self.app.put_json('/models/%s' % self.model_id,
+                          {'definition': self.valid_definition},
+                          headers=self.headers)
+        self.app.patch_json('/models/%s/acls' % self.model_id,
+                            {"admin": ["-ALL"]},
```
almet, line 121: should it be in upper case or is it just case insensitive?

daybed/tests/test_views.py
```diff
@@ -102,6 +136,10 @@ def test_spore_get(self):
 
 class ModelsViewsTest(BaseWebTest):
 
+    def __init__(self, *args, **kwargs):
+        self.maxDiff = None
```
almet, line 140: I actually like the fact that it doesn't clutter my output when tests are wrong, nope?

daybed/tests/test_views.py
```diff
@@ -148,21 +186,190 @@ def test_cors_support_on_404(self):
...
         self.assertDictEqual(resp.json, acls)
 
+    def test_patch_acls_add(self):
+        self.app.put_json('/models/test',
+                          MODEL_DEFINITION,
+                          headers=self.headers)
+        try:
+            self.db.add_token('alexis', 'bar')
+            self.db.add_token('remy', 'foobar')
+        except TokenAlreadyExist:
```
almet: why do we need the except here?

daybed/tests/test_views.py
Natim, line 140: In case of DictEqual for this particular tests you want to see the all Dict diff. It is useful when test broke but nothing will be broken if test are ok.

daybed/tests/test_views.py
```diff
@@ -148,21 +186,190 @@ def test_cors_support_on_404(self):
...
+        acls[u"admin"].remove("read_acls")
+        acls[u"admin"].remove("update_acls")
+        self.assertDictEqual(resp.json, acls)
+
+    def test_patch_acls_add_all(self):
+        self.app.put_json('/models/test',
+                          MODEL_DEFINITION,
+                          headers=self.headers)
+        try:
+            self.db.add_token('alexis', 'bar')
```
almet: in the setup / teardown functions we should add the tokens and remove them, nope? (rather than doing that for every test)

daybed/tests/test_views.py
Natim, line 121: It is case insensitive but I like the fact of using ALL to differ this shortcut from standard permissions

daybed/tests/test_views.py
Natim: To not break tests if they already exists.

daybed/tests/test_views.py
Natim: We don't do it for every test. We are doing it only if we need them.

almet: We're missing the scoped beforeEach / afterEach of mocha ;)

Natim: Yes :+1:

daybed/views/errors.py
```diff
@@ -1,7 +1,16 @@
...
 
-def unauthorized_view(request):
-    return Response('{"error": "401 Unauthorized",'
-                    ' "msg": "You must login to access this page."}',
-                    status='401 Unauthorized', content_type='application/json')
+def forbidden_view(request):
+    if not request.token or request.token == Everyone:
+        return Response(
+            '{"error": "401 Unauthorized",'
+            ' "msg": "You must login to access this page."}',
```
almet: "You must be logged-in to access this page"

daybed/views/errors.py
```diff
@@ -1,7 +1,16 @@
...
+            status='401 Forbidden', content_type='application/json')
```
almet: "401 Forbidden" sounds wrong to me.

daybed/views/errors.py
```diff
@@ -1,7 +1,16 @@
...
+    return Response(
+        '{"error": "403 Forbidden",'
+        ' "token": "%s", '
+        ' "msg": "You don\'t have the required ACLs '
```
almet: I would change this as "Access to this resource is Forbidden". Also, you don't need to escape the `'`

daybed/views/fields.py
```diff
@@ -6,8 +6,7 @@
 
 fields = Service(name='fields',
                  path='/fields',
-                 description='The resource containing all the fields',
-                 renderer='jsonp')
```
almet, line 10: why not jsonp?

Natim, line 10: Because cornice doesn't seems to use this renderer value and that tests are still passing without it.

daybed/views/models.py
```diff
@@ -50,7 +58,57 @@ def get_acls(request):
...
+    """Update a model acls."""
+    model_id = request.matchdict['model_id']
+    definition = request.db.get_model_definition(model_id)
+    acls = dict_list2set(request.db.get_model_acls(model_id))
+
+    for token, perms in iteritems(request.validated['acls']):
+        # Handle remove all
+        if '-all' in [perm.lower() for perm in perms]:
+            for p in PERMISSIONS_SET:
+                acls[p].discard(token)
```
almet: I love sets.

daybed/views/fields.py
Natim, line 10: Also CORS has been added so we shouldn't need jsonp anymore.

daybed/views/models.py
```diff
@@ -50,7 +58,57 @@ def get_acls(request):
...
+def patch_acls(request):
+    """Update a model acls."""
+    model_id = request.matchdict['model_id']
+    definition = request.db.get_model_definition(model_id)
+    acls = dict_list2set(request.db.get_model_acls(model_id))
+
+    for token, perms in iteritems(request.validated['acls']):
+        # Handle remove all
+        if '-all' in [perm.lower() for perm in perms]:
+            for p in PERMISSIONS_SET:
```
almet: I would rename `p` to `perm`

daybed/views/models.py
```diff
@@ -82,7 +140,11 @@ def delete_model(request):
         model = request.db.delete_model(model_id)
     except ModelNotFound:
         request.response.status = "404 Not Found"
-        return {"msg": "%s: model not found" % model_id}
+        return {
```
almet: I think we can even do something with cornice like that:

``` python
request.errors.add('model_id', 'querystring', 'model does not exist')
request.errors.status_code = 404
```

Or something similar, I don't qiute remember the details, but that's possible ;)

Natim: `s/querystring/path`

daybed/views/records.py
```diff
@@ -83,12 +92,19 @@ def get(request):
...
+            "error": "404 Not Found",
+            "msg": "%s: record not found %s" % (model_id, record_id)
+        }
 
 
 @record.put(validators=record_validator, permission='put_record')
 def put(request):
     """Updates or creates a record."""
+    # if we are asked only for validation, don't do anything more.
+    if request.headers.get('Daybed-Validate-Only', 'false') == 'true':
```
almet: Not related to this PR, but I guess we should remove the "Daybed" prefix :)

almet: ref https://github.com/spiral-project/daybed/issues/187

docs/source/permissions.rst
```diff
@@ -1,211 +1,57 @@
...
 In *Daybed*, permissions allow to control what are the privileges on
-models, records, users ...and permissions !
+models, records and acls !
 
 They allow to express the following strategies :
 
-- "Anyone can create new records on this model"
+- "Everyone can create new records on this model"
 - "Alexis is able to delete records created by others"
-- "Users of the group 'rangers' have the right to create new records"
```
almet, line 11: Why removing this one?

Natim, line 11: There is no user nor group anymore.

daybed/views/models.py
almet: `request.errors.status = 404`

docs/source/permissions.rst
```diff
@@ -1,211 +1,57 @@
...
-Roles are defined on the model as a name and a list of users or groups.
-Custom roles can be defined, but some are provided by default :
-
-* ``role:admins``
-* ``role:authors``
-
-:notes:
-
-    The creator of a **model** is put among the list of users for the ``admins`` role.
+There is a list of 12 permissions that you can add to a token on a
```
almet: I think we shouldn't put a number because that's likely that we'll add new ones and forget to increase it ;)

almet: (plus who cares how many there are? :)

docs/source/permissions.rst
```diff
@@ -1,211 +1,57 @@
...
+- **delete_model**: You can delete a model
+- **create_record**: You can add an entry to the model
+- **read_all_records**: You can read all model's records
+- **update_all_records**: You can update all model's records
+- **delete_all_records**: You can delete any model's records
+- **read_own_records**: You can read records on which you are an author
+- **update_own_records**: You can update and change records on which you are an author
+- **delete_own_records**: You can delete records on which you are an author
 
+These are permissions that you can set on a model.
```
almet: "these are the permissions you can define on a model"

docs/source/permissions.rst
```diff
@@ -1,211 +1,57 @@
...
-        'type': 'data',
-        'authors': ['john'],
-        'model_id': 'todo',
-        'data': {"item": "finish the documentation", "status": "todo"}
-    }
+Global permissions
+==================
 
-Of the following model:
+There is three more permissions that are configured at the server level:
```
almet: There are*

docs/source/permissions.rst
```diff
@@ -1,211 +1,57 @@
...
 
-.. code-block::
+- **create_model**: List of people that have the right to create a model
```
almet: list of tokens which allow to create models

docs/source/permissions.rst
```diff
@@ -1,211 +1,57 @@
...
+- **create_token**: List of people that can create tokens
```
almet: s/people/tokens

docs/source/permissions.rst
```diff
@@ -1,211 +1,57 @@
...
+- **manage_tokens**: List of people that have the right to delete tokens
```
almet: same.

docs/source/permissions.rst
```diff
@@ -1,211 +1,57 @@
...
-policy      -R--
-roles       -R--
-==========  ========
-
-Hence, he will not have the permission to modify it.
-
-Alexis is in the `admins` group, which is given `admins` role on
-this model. If he tries to modify the record, he will obtain full permissions.
+At the API level, you will often need more than one permissions to get
+access to a page.
```
almet: I don't understand what this sentence tries to tell.

almet: That's a bit clearer with the next paragraph. But what's a page?

Natim: API endpoint.

docs/source/usage.rst
```diff
@@ -286,7 +302,196 @@ Get back a definition
...
 
-XXX Todo
+Get back the model ACLs
+-----------------------
+
+**GET /models/{modelname}/acls**
+
+::
+
+    http GET http://localhost:8000/models/todo/acls --verbose --auth-type=hawk --auth='504fd8148d7cdca10baa3c5208b63dc9e13cad1387222550950810a7bdd72d2c:'
```
almet: I've tried to keep the --auth and hawk things on a new line to be easier to read:

```
http GET http://localhost:8000/models/todo/acls --verbose \
--auth-type=hawk  --auth='504fd8148d7cdca10baa3c5208b63dc9e13cad1387222550950810a7bdd72d2c:'
```

docs/source/usage.rst
```diff
@@ -286,7 +302,196 @@ Get back a definition
...
+        ]
+    }
+
+
+Add some rights
+---------------
+
+You can add rights to an existing token, Authenticated people or Everyone.
+
+Pyramid define two principals: system.Authenticated and system.Everyone
```
almet: We shouldn't talk about "principals" here, nor "Pyramid". The users don't care.

docs/source/usage.rst
```diff
@@ -286,7 +302,196 @@ Get back a definition
...
+`update_acls` to alexis we would write::
+
+    {
+        "Authenticated": ["read_definition", "read_acls"],
+        "alexis": ["-update_acls"]
+    }
+
+For this to be valid, `alexis` must be an existing token.
+
+If you want to add or remove all the right to/from somebody, you can use the shortcut ALL::
```
almet: `you can use the `ALL` shortcut.`

docs/source/usage.rst
```diff
@@ -286,7 +302,196 @@ Get back a definition
...
+
+    {
+        "Authenticated": ["-ALL"],
+        "alexis": ["+ALL"]
+    }
+
+If you don't provide the `-` or the `+` in front of the right we assume a `+`.
```
almet: "we assume it means you want to add the permission".

almet: We shouldn't name these "rights" but "permissions" to keep consistent.

docs/source/usage.rst
```diff
@@ -286,7 +302,196 @@ Get back a definition
...
+        "Authenticated": ["all"]
+    }
+
+Is equivalent to::
+
+    {
+        "Authenticated": ["+all"]
+    }
+
+You can write rights both in lower or uppercase.
```
almet: I believe we should ask people to send it in lower case, not upper case (and even accept upper case values, but not tell it to anyone).

almet: ("There should be one obvious way to do things")

docs/source/usage.rst
```diff
@@ -286,7 +302,196 @@ Get back a definition
...
+
+In case you try to add a non existing right, or to modify right of a
+non existing token, you will get an error.
+
+If you need to remove rights from a removed token, you will have to use the PUT endpoint.
```
almet: Remove rights from a removed token? What would that mean?

docs/source/usage.rst
```diff
@@ -286,7 +302,196 @@ Get back a definition
...
+
+**PATCH /models/{modelname}/acls**
+
+::
+
+   echo '{"Everyone": ["read_definition"]}' | http PATCH http://localhost:8000/models/todo/acls \
```
almet: Ah, cool, wasn't aware we could use httpie like that.

docs/source/usage.rst
```diff
@@ -286,7 +302,196 @@ Get back a definition
...
+            "update_own_records"
+        ],
+        "system.Everyone": [
+            "read_definition"
+        ]
+    }
+
+**PUT /models/{modelname}/acls**
+
+This is means to replace ACLs for a model. It could be useful in case
```
almet: "This endpoint lets you replace a set of ACLs for a model"

docs/source/usage.rst
```diff
@@ -286,7 +302,196 @@ Get back a definition
...
+where PATCH doesn't work (remove rights for a removed token.) or to
+replace all ACLs in one call.
```
almet: "to replace all the permissions"

almet: I even wonder if we wouldn't like to call the endpoint `{PUT,GET,PATCH} /models/{modelname}/permissions` (rather than `acls`)

daybed/tests/test_acl.py
almet, line 80: yup but they could be useful to someone else in the future :)

daybed/views/fields.py
almet, line 10: true fact :)

docs/source/permissions.rst
almet, line 11: ah right :dancer:

docs/source/permissions.rst
almet: okay, then we should replace that :)

almet:
That overall is pretty exciting, thanks for the awesome work on all this!

docs/source/usage.rst
Natim: It means the token doesn't exists anymore but as some permissions set for it on this model.

docs/source/usage.rst
Natim: Yes probably

almet:
merged!

Issue #192, opened 2014-08-07
Add test suite for records of a group field


leplatrem:
See https://github.com/spiral-project/daybed/issues/189#issuecomment-51542743

Issue #189, opened 2014-08-04
Storing a validating a List of items


Natim:
I'd like to be able to store and validate something like so:

```
[
  {"id": 1, "type": "text", "title": "This is a Text entry", "text": "Hello World."},
  {"id": 2, "type": "text", "title": "Lorem ipsum dolor sit amet",
   "text": "**Consectetur adipisicing elit**, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua."},
  {"id": 3, "type": "todo", "title": "Today", "tasks": [
    {"label": "Walk the dog", "done": true},
    {"label": "Clean the car", "done": false},
    {"label": "Buy a new carpet", "done": false}
  ]},
  {"id": 4, "type": "text", "title": "Ut enim ad minim veniam",
   "text": "Quis nostrud *exercitation ullamco* laboris nisi ut aliquip ex ea commodo consequat."},
  {"id": 5, "type": "text", "title": "Duis aute irure",
   "text": "> Dolor in [reprehenderit](http://google.com/) in voluptate velit esse cillum dolore eu fugiat nulla pariatur."},
  {"id": 6, "type": "text", "title": "Excepteur sint occaecat",
   "text": "    Cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."},
  {"id": 7, "type": "todo", "title": "Trip to Paris", "tasks": [
    {"label": "Eat cheese", "done": false},
    {"label": "Eat more cheese", "done": false},
    {"label": "See Eiffel Tower", "done": false},
    {"label": "MOAR CHEESE", "done": false}
  ]}
]
```

In other words the definition it'd like would looks like this:

```
{
    "title": "kept",
    "description": "A list of items I need to keep track on",
    "fields": [
        {
            "name": "title",
            "type": "string",
            "label": "Title"
        },
        {
            "name": "type",
            "type": "enum",
            "label": "Type",
            "choices": ["text", "todo", "picture", "video"]
        },
        {
            "name": "tasks",
            "type": "list",
            "definition": {
                {
                  "name": "label",
                  "type": "string",
                  "label": "Label"
                },
                {
                    "name": "done",
                    "type": "boolean",
                    "label": "Done"
                }
            },
            "label": "Tasks",
            "linkedTo": {"type": "todo"}
        },
        {
            "name": "text",
            "type": "string",
            "label": "Text",
            "linkedTo": {"type": "text"}
        }
     ]
}
```

[...]

Issue #88, opened 2013-10-11
How to handle multiple fields data ? (Date, RIB, Full name, address)


Natim:
Is it only a Backbone problem or also a daybed one ?

leplatrem:
Duplicate of #11  :)

Issue #11, opened 2012-06-01
Groups of fields with title and description


leplatrem:
At model creation, it could be possible to define groups of fields:

```
SCHEMA = {
    name: "My super event",
    description: "",
    fields: [
        {   type: "string",
            name: "title",
            description: "blah",
        },
        {   name: "category",
            type: "enum",
            description: "blah",
            choices: ['sport', 'culture']
        },
        {   name: "street",
            type: "string",
        },
        {   name: "city",
            type: "string",
        },
    ],
    groups: [
        {  title: "Home",
           description:  "Your address helps us find you",
           fields: ["street", "city"]
        },
    ]
}
```

This would allow the UI to nicely layout the form for this model. 

Since groups would be optional, they do not interfere with existing models, or those who don't need UI forms.

PR #193, opened 2014-08-08
Fix group field validation

leplatrem:
~~I'm stuck with group fields, and I'm currently thinking of getting rid of them.~~

My first idea was just to allow grouping of fields, in a way that does not affect their individual validation at all.

For example, the following model : 

``` javascript
{name: 'person',
 fields: [
    {name: 'age', type: 'int'},
    {description: 'address',
     type: 'group',
     fields: [{name: 'street', type: 'string'},
              {name: 'city', type: 'string'}]
    }
 ]
}
```

could validate this record

``` javascript
{age: 12, street: 'bakery', city: 'london'}
```

I must admit that the group field type I had in mind belongs more to the presentation layer (grouping widgets in forms) than the model layer.

The current implementation is completely wrong : it does not validate at all ! (see #189, #192)

I tried to hack around, ~~but I cannot find~~ I found a way to flatten the fields list (i.e. expending the fields at the model level with the ones from the group fields).

~~And if we validate in a dict (e.g. `{age: 12, '': {street: 'bakery', city: 'london'}}`), then there is no difference with the `object` field, which is well tested and rather intuitive...~~

~~Any idea ? Opposition to `group` field type removal ?~~

### DIFF
--- a/daybed/schemas/__init__.py
+++ b/daybed/schemas/__init__.py
@@ -79,7 +79,7 @@ class TypeField(object):
     hint = u''
 
     @classmethod
-    def definition(cls):
+    def definition(cls, **kwargs):
         schema = SchemaNode(Mapping())
         schema.add(SchemaNode(String(), name='name',
                               validator=Regex(r'^[a-zA-Z][a-zA-Z0-9_\-]*$')))
@@ -104,10 +104,11 @@ class TypeField(object):
 
 class TypeFieldNode(SchemaType):
     def deserialize(self, node, cstruct=null):
+        kwargs = dict(node=node, cstruct=cstruct)
         try:
-            schema = registry.definition(cstruct.get('type'))
+            schema = registry.definition(cstruct.get('type'), **kwargs)
         except UnknownFieldTypeError:
-            schema = TypeField.definition()
+            schema = TypeField.definition(**kwargs)
         return schema.deserialize(cstruct)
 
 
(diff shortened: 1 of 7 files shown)


almet:
Hmm, interesting. How one could test that the items of a list follow some semantics (for instance that they're integers)?

I may not understand completely how all this work, and I think that having a full fledged documentation about how do deal with this would be useful.

Also (might be related), I'm currently playing a bit with jsonschema, we might take some of their logic. See https://spacetelescope.github.io/understanding-json-schema/reference/array.html for instance.

leplatrem:
What you mention here (a list of integers) was implemented in https://github.com/spiral-project/daybed/pull/161
and it has nothing to do with this PR :)

This one is about fixing the validation of groups of fields (see #11, #88)

Natim:
All I am asking is please don't remove the group fields. We are using it for n1k0/kept and I was so happy to find it. I think we should do recursive validation for this and handle group field as if it was an anonymous model inside a model

leplatrem:
You were happy to find it because it was the only field type whose validation was not working. See #192.

What you need for kept is `object`, and it already does recursive validation of anonymous models!

Natim:
Can you please give me the model definition you are taking about?
Le 9 août 2014 09:36, "Mathieu Leplatre" notifications@github.com a écrit
:

leplatrem:
I merge and close this, since this PR is only a fix on existing code and should not introduce more confusion.

I will take the time to write some documentation of the fields.

@Natim the model for kept can be something like this : 

``` javascript
   'tasks': {
       'type': 'list',
       'itemtype': 'object',
       'parameters': {
            'fields': [
                {'type': u'enum',
                 'name': u'status',
                 'choices': [u'todo', u'done']}
            ]
       }
   }
```

PR #194, opened 2014-08-08
Rename ACL to permission

leplatrem:
Basically everywhere...


almet:
Yes, definitely.

Natim:
Please also change daybed.js accordingly
Le 9 août 2014 01:48, "Alexis Metaireau" notifications@github.com a écrit
:

PR #195, opened 2014-08-09
[WIP] Add Json schema support

almet:
- [x] Plug the new renderer with cornice;
- [x] Test that all the fields are supported;
- [x] Decide on a mime type to use.
- [ ] Accept JSON schema for definitions.

Here are the types supported by Daybed that we need to convert to JSON Schema:
- [x] Integer
- [x] Text
- [x] Datetime
- [x] Boolean
- [x] Regex
- [ ] JSON
- [ ] Anyof
- [x] Email
- [ ] OneOf
- [x] String
- [ ] Object
- [ ] Date
- [x] Url
- [ ] List
- [ ] Choices
- [x] Range

GeoTypes:
- [ ] Point
- [ ] Polygon
- [ ] GeoJSON
- [ ] Line

### DIFF
--- a/daybed/__init__.py
+++ b/daybed/__init__.py
@@ -24,7 +24,7 @@ from daybed.acl import (
     RootFactory, DaybedAuthorizationPolicy, check_api_token,
 )
 from daybed.views.errors import forbidden_view
-from daybed.renderers import GeoJSON
+from daybed.renderers import GeoJSON, JSONSchema
 from daybed.backends.exceptions import TokenNotFound
 
 
@@ -108,6 +108,7 @@ def main(global_config, **settings):
     config.add_subscriber(add_default_accept, NewRequest)
 
     config.add_renderer('jsonp', JSONP(param_name='callback'))
-
     config.add_renderer('geojson', GeoJSON())
+    config.add_renderer('jsonschema', JSONSchema())
+
     return config.make_wsgi_app()
--- a/daybed/renderers.py
+++ b/daybed/renderers.py
@@ -21,6 +21,7 @@ class JSONSchema(JSONP):
             'url': lambda x: {'type': 'string', 'format': 'uri'},
             'enum': self.serialize_enum,
             'range': self.serialize_range,
+            'datetime': self.serialize_datetime,
         }
         super(JSONSchema, self).__init__(*args, **kwargs)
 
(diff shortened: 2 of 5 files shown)


leplatrem:
Great ! We should definitely follow standards !  ref #53
I could help you on this, this is quite a big task :)

almet:
I added some changes to the branch, and I updated the description to contain the list of fields we currently support in daybed.

I'll be out starting tomorrow for one week, don't hesitate to continue what I started :)

almet:
A few outstanding questions:
- How do we deal with relation between models? (I guess we should set them to integers / strings depending the case);
- How do we handle the JSON and GeoJSON fields? We could do a [(really long) regexp](https://stackoverflow.com/questions/2583472/regex-to-validate-json), but I wonder if there is another take on it?
- Handling objects and lists will be fun :)

almet:
FWIW, here is the doc I'm using as a reference https://spacetelescope.github.io/understanding-json-schema/

Natim:
For relations between models modelname is a string.

leplatrem:
I was wondering if it is still relevant to have our own schema formalism... I mean, instead of formatting our custom stuff to the standard JSON schema, just accept it as input for model creation ?

Natim:
For know it is relevant because our schema is close to what daybed do.
Btw we could implement both input formalisms if we need to be compatible somehow and ask the question again then.

almet:
@leplatrem, I was wondering the exact same thing.

I believe what we have for daybed is simpler than json schema, but I may be wrong. Especially, for the geometric fields, there is nothing that deals with them in json schema, so that could be a bit hard to do.

For really simple cases, I believe we should accept JSON Schema definitions, and that would make us open to a list of services that already support this formalism.

However, that's in my opinion a separate issue. In this one, I think we should focus on outputing a json schema from a daybed (colander) definition.

almet:
for now, mimetype is `application/schema+json`

almet:
@leplatrem do you have any idea how we can deal with geographic fields here?

leplatrem:
I found only one example, and it looks pretty complete : https://github.com/fge/sample-json-schemas/blob/master/geojson/geometry.json 

In Daybed we have a geojson geometry field (as above), and some basic geo fields that could be described with the positions part only (coordinates only)

PR #198, opened 2014-08-19
ElasticSearch integration

leplatrem:
Revamp of  https://github.com/spiral-project/daybed/pull/173

Ready to merge or review !


conf/development.ini
```diff
@@ -2,6 +2,7 @@
 use = egg:daybed
 # Backend configuration
 daybed.backend = daybed.backends.redis.RedisBackend
+index.host = localhost:9200
```
almet: should we name this `elasticsearch.host`? I think users would know better what this configuration is about if we word it with "elasticsearch" in it.

daybed/__init__.py
```diff
@@ -99,6 +108,16 @@ def add_db_to_request(event):
 
     config.add_subscriber(add_db_to_request, NewRequest)
 
+    # index initialization
+    index_hosts = build_list(settings.get('index.host', 'localhost:9200'))
```
almet: Why do we need to have a list here? Is it the way elastisearch handles the sharding?

daybed/indexer.py
```diff
@@ -0,0 +1,166 @@
...
+        config.add_subscriber(self._update_index, events.ModelUpdated)
+        config.add_subscriber(self._delete_index, events.ModelDeleted)
+        config.add_subscriber(self._index_record, events.RecordCreated)
+        config.add_subscriber(self._reindex_record, events.RecordUpdated)
+        config.add_subscriber(self._unindex_record, events.RecordDeleted)
+
+    def search(self, model_id, query, params):
+        supported_params = ['sort', 'from', 'source', 'fields']
+        params = dict([p for p in params.items() if p[0] in supported_params])
+        return self.client.search(index=model_id,
```
almet, line 17: That's a matter of taste, but I think I would feel more confortable if we have these defined on a new line, like that:

``` python
return self.client.search(
    index=model_id,
    doc_type=model_id,
    body=query,
    **params
)
```

Or maybe we could expend all this on two lines rather than 4, maxing at 80chars.

daybed/indexer.py
```diff
@@ -0,0 +1,166 @@
...
+
+from daybed import events, logger
+
+
+class DaybedIndexer(object):
+
+    def __init__(self, config, hosts):
+        self.client = elasticsearch.Elasticsearch(hosts)
+
+        config.add_subscriber(self._create_index, events.ModelCreated)
```
almet: Interestingly, I would expect to have these methods to be public in this class, and the binding with the events to occur in a different place (maybe in a function in this module, but that's IMO not related to the `DaybedIndexer` class itself).

daybed/__init__.py
Natim: ES handle the sharding but you need to load-balance requests as well so that your all cluster is used if not you slow down the engine because one bit is doing all the work (request other nodes and agregate the results)

conf/development.ini
Natim: :+1:

daybed/indexer.py
```diff
@@ -0,0 +1,166 @@
+import json
+
+import elasticsearch
+from elasticsearch.exceptions import ElasticsearchException
+
+from daybed import events, logger
+
+
+class DaybedIndexer(object):
```
almet: Should we name this `ElasticSearchIndexer`? We know it's a part of daybed already.

daybed/indexer.py
Natim, line 17: You are doing too much JavaScript these days. :dart:

daybed/indexer.py
```diff
@@ -0,0 +1,166 @@
...
+                                                          event.model_id))
+        try:
+            self.client.delete(index=event.model_id,
+                               doc_type=event.model_id,
+                               id=event.record_id,
+                               refresh=True)
+        except ElasticsearchException as e:
+            logger.error(e)
+
+    def __put_mapping(self, request, model_id):
```
almet: I would have expected this to take the information it needs in the parameters, rather than unfolding them from the request.

But (yes, I think as I type) this seem to come at the cost of complexifying the code, just for the sake of a better separation of concerns.

If you think it's better to have this class completely tied to the notion of requests and the notion of events, then I think we should go for it. If the separation of concerns is easy to implement, maybe we should go for it :)

conf/development.ini
leplatrem: Yeah, naively I first thought there could be many index engines :)

daybed/indexer.py
leplatrem: :+1:

daybed/indexer.py
almet, line 17: So true.

almet:
Cool! Thanks :)

PR #203, opened 2014-08-29
Hawk review concepts

leplatrem:
- [x] Clarify usage of token versus of identifier
- [x] Review and improved doc
- [x] Reformat of current credentials in hello (?..maybe wrong to mention credentials)
- [ ] Complete glossary

For this pull request, its author has lost 12 points of energy.


daybed/__init__.py
```diff
@@ -21,18 +21,20 @@
...
 )
 from daybed.views.errors import forbidden_view
 from daybed.renderers import GeoJSON
-from daybed.backends.exceptions import TokenNotFound
+from daybed.backends import exceptions as backend_exceptions
 from daybed import indexer, events
 
 
-def get_hawk_id(request, tokenid):
+def get_user_credentials(request, identifier):
```
Natim, line 32: Please avoid using user, we removed it, use token instead. also hawk_id is more specific because it is the derived id from the token.

Natim, line 32: `hkdf(SessionToken, 32*2)` -> `credentials` > `tokenId = credentials.id; authKey = credentials.key`

daybed/__init__.py
```diff
@@ -21,18 +21,20 @@
...
+    """Lookup the identifier, and get its secret key.
+    """
     try:
-        return tokenid, request.db.get_token(tokenid)
-    except TokenNotFound:
+        return identifier, request.db.get_secret_key(identifier)
```
Natim: I am currently changing `get_token` to return the `sessionToken` rather than the `secret_key` in another pull-request.

daybed/__init__.py
```diff
@@ -21,18 +21,20 @@
...
+    except backend_exceptions.IdentifierNotFound:
```
Natim, line 37: `Identifier` is confusing I think because `Identifier` means `id`, here we are talking about `Token` from its `id` to be more specific.

Natim:
Please don't change the `Token` -> `Identifier` wording.
The aims is to keep Daybed simple by just handling secure tokens generated on server side.

If you change to identifier people can decide to use an `email identifier` with a `password` secret instead of tokens and we should enforce this will not happen for security reason and simplicity.

Also we will use `sessionToken` directly in the URL for some HTML5 apps (like the Todo one) so I wanted change `get_token(tokenId)` to return the `sessionToken` rather than only the `authKey` so we can then always return the sessionToken on the oauth endpoint.

almet:
Not valid anymore, closing.

PR #208, opened 2014-08-29
Add a delete_indices helper to delete all indices during tests.

Natim:


conf/tests.ini
```diff
@@ -11,6 +10,9 @@ elasticsearch.hosts = localhost:9200
 # backend.db_host = http://localhost:5984/
 # backend.db_name = daybed-test
 
+elasticsearch.hosts = localhost:9200
+elasticsearch.indices_prefix = daybed_tests
```
leplatrem, line 14: not necessary I guess (a lot of default values everywhere)

conf/development.ini
```diff
@@ -3,6 +3,7 @@ use = egg:daybed
 # Backend configuration
 daybed.backend = daybed.backends.redis.RedisBackend
 elasticsearch.hosts = localhost:9200
+elasticsearch.indices_prefix = daybed
```
leplatrem: not necessary (default value)

daybed/indexer.py
```diff
@@ -8,21 +8,22 @@
 
 class ElasticSearchIndexer(object):
 
-    def __init__(self, hosts):
+    def __init__(self, hosts, prefix='daybed'):
```
leplatrem: no default necessary here

conf/development.ini
Natim: ok

conf/tests.ini
Natim, line 14: We shouldn't use the default for tests.

conf/development.ini
Natim: Also I think if people uses it as a sample, they should be aware of this `indices_prefix`

Natim:
Updated

PR #206, opened 2014-08-29
Improve Token, Credentials wording + store both credentials_key and token.

Natim:

PR #210, opened 2014-08-29
GET /token return current authorization credentials

Natim:


Natim:
Based on #206 should be merged before.

PR #216, opened 2014-09-01
Use the version as a prefix in the API

almet:


almet:
Hmm, pypy! Trying to run again.

PR #218, opened 2014-09-01
Prevent ES errors if not available

leplatrem:


almet:
We should add a test for this, mocking the connection to ES and making it fail to check no exception is raised.

leplatrem:
:+1: (I mean, +1)

leplatrem:
updated !

PR #219, opened 2014-09-01
Change the way to define lists (fixes #217)

### LINKED ISSUE #217: List item type definition is cumbersome
https://github.com/n1k0/kept/pull/25/files#r16965400

leplatrem:
See #217 

Usage is easier. Code is shorter. What else?™

### DIFF
--- a/daybed/schemas/list.py
+++ b/daybed/schemas/list.py
@@ -1,29 +1,14 @@
 from pyramid.i18n import TranslationString as _
-from colander import (SchemaNode, String, Sequence, drop, OneOf,
-                      Invalid, Mapping)
+from colander import SchemaNode, Sequence, drop
 
+from . import TypeFieldNode
 from .base import registry, TypeField, JSONList
 
 
-class ParametersMatchItemType(object):
-    def __call__(self, node, value):
-        if 'parameters' in value and 'itemtype' not in value:
-            msg = u'Missing itemtype since parameters is provided'
-            raise Invalid(node, msg)
-        if 'itemtype' in value and 'parameters' in value:
-            itemtype = value['itemtype']
-            parameters = value['parameters']
-            schema = registry.definition(itemtype)
-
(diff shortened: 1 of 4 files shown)


daybed/tests/test_schemas_list.py
```diff
@@ -18,15 +19,18 @@ class ListFieldTest(unittest.TestCase):
...
     def test_can_be_defined_with_a_field_type(self):
         field = self.schema.deserialize(self.definition)
+        for defaultattr in ['required', 'label']:
+            field.pop(defaultattr)
+            field['item'].pop(defaultattr)
         self.assertDictEqual(self.definition, field)
 
     def test_is_not_valid_if_field_type_unknown(self):
-        self.definition['itemtype'] = 'asteroid'
+        self.definition['item']['type'] = 'asteroid'
```
almet, line 33: `']['` ftw.

almet:
Sounds good to me.

This would need to have the documentation updated as well.

daybed/schemas/list.py
```diff
@@ -38,20 +23,13 @@ class ListField(TypeField):
...
         schema = super(ListField, cls).definition(**kwargs)
-        schema.add(SchemaNode(String(),
-                              name='itemtype',
-                              validator=OneOf(registry.names),
-                              missing=drop))
-        schema.add(SchemaNode(Mapping(unknown='preserve'),
-                              name='parameters',
-                              missing=drop))
-        schema.validator = ParametersMatchItemType()
+        schema.add(SchemaNode(TypeFieldNode(named=False),
```
almet, line 26: Okay, so it's only named when it's the top level object?

leplatrem, line 26: Yes, it's only named when used in a definition fields list

leplatrem:
Which documentation ? :)

almet:
We should have some doc about all the existing types and how to use them.

PR #220, opened 2014-09-03
Minor docs update

leplatrem:


docs/source/permissions.rst
```diff
@@ -56,3 +46,218 @@ For example, if you want to get a complete model (definition and records),
...
+            "update_definition",
+            "update_own_records"
+            "update_permissions",
+        ]
+    }
+
+Modification
+------------
+
+We use ``-`` and ``+`` to modify the an existing set of permissions for an
```
Natim: the an

docs/source/permissions.rst
```diff
@@ -56,3 +46,218 @@ For example, if you want to get a complete model (definition and records),
...
+idenfier.
```
Natim: identifier

docs/source/permissions.rst
```diff
@@ -56,3 +46,218 @@ For example, if you want to get a complete model (definition and records),
...
+            "update_all_records",
+            "delete_all_records"
+        ],
+        "220a1c..871": [
+            "ALL",
+        ]
+    }
+
+
+If the *administration* wants to share her privileges with other, she can either :
```
Natim: - administration -> administrator
- no space before `:`

docs/source/permissions.rst
```diff
@@ -56,3 +46,218 @@ For example, if you want to get a complete model (definition and records),
...
+    }
+
+
+Online poll
+-----------
+
+Everybody can answer the poll, but are not allowed to correct their answers,
+nor to see the poll results.
+
+We keep ``read_definition`` as it might used to build the form on the client-side :
```
Natim: it might be used.

docs/source/permissions.rst
```diff
@@ -56,3 +46,218 @@ For example, if you want to get a complete model (definition and records),
...
+- **manage_tokens**: List of identifiers allowed to delete tokens
+
+
+Usage
+=====
+
+Permissions are set on models, as a *dictionnary* between :term:`identifiers` (*key id*)
+and lists of :term:`permissions`.
+
+We refer to the whole world with the special *id* ``Everyone`` (or ``system.Everyone``)
```
almet: I believe we shouldn't expose the users with the `system` prefix, one id for once concept is better than two, and the clients should always be given `Everyone` and `Authenticated`.

Natim: Yes we had the exact same reflexion with @leplatrem

docs/source/permissions.rst
```diff
@@ -56,3 +46,218 @@ For example, if you want to get a complete model (definition and records),
...
+and to authenticated users with ``Authenticated`` (or ``system.Authenticated``).
+
+:notes:
+
+    With Hawk, the *key ids* look like :term:`tokens`, but the *id* is the part of
```
almet: We should add a link to some documentation about Hawk, since this page doesn't refer to it already. Ideally it should be a link on daybed docs on how to use Hawk.

docs/source/permissions.rst
Natim: I have created https://github.com/spiral-project/daybed/issues/233 about it.

docs/source/permissions.rst
```diff
@@ -56,3 +46,218 @@ For example, if you want to get a complete model (definition and records),
...
+    your :term:`credentials`, and is different from the session token.
+
+
+When you create a model, you gain the full set of available permissions.
+
+This means that the credentials id you used in the request will be associated to all permissions ::
```
almet: I would rename "The credentials id" with "the identifier".

docs/source/permissions.rst
```diff
@@ -56,3 +46,218 @@ For example, if you want to get a complete model (definition and records),
...
+            "read_permissions",
+            "update_all_records",
+            "update_definition",
+            "update_own_records"
+            "update_permissions",
+        ]
+    }
+
+
+Let's say, you want to allow authenticated users to create records and manage
```
almet: You don't need to `,` here, it interrupts a bit the reading flow.

docs/source/permissions.rst
```diff
@@ -56,3 +46,218 @@ For example, if you want to get a complete model (definition and records),
...
+            "update_definition",
+            "update_own_records"
+            "update_permissions",
+        ]
+    }
+
+Modification
+------------
+
+We use ``-`` and ``+`` to modify the existing set of permissions for an
```
almet: We're mixing `we` and `you` in this document. We should always use "you". 

This sentence then becomes:

> "You can use `-` and `+` to modify the existing set of permissions for an identifier"

docs/source/permissions.rst
```diff
@@ -56,3 +46,218 @@ For example, if you want to get a complete model (definition and records),
...
+identifier.
+
+To grant `create_record` to anonymous users, ``read_permissions`` to
+authenticated users and remove `update_permissions` from *id* ``220a1c..871``
+we would send the following request ::
```
almet: ditto with the "we" / "you".

almet:
I added a couple of comments but it's generally looking very good.

In general, it seems that we're mixing "we" and "you" when talking about the user, in the documentation, we should decide what's best and keep consistent.

almet:
Thanks!

PR #221, opened 2014-09-03
WIP - Document field types

leplatrem:


almet:
Awesome, thanks!

almet:
We should also describe what they do with an example, but let's keep that for another bug ;)

Issue #48, opened 2013-02-14
CORS doesn't work on 404ed resources


almet:
when resources don't exist, we should return the right CORS headers. I'm not sure we're doing this ATM.

leplatrem:
It looks like this applies to 400 too.

Wouldn't it be a Cornice bug instead ?

leplatrem:
BTW, we have two ways of raising 404 :

https://github.com/spiral-project/daybed/blob/master/daybed/validators.py#L36
https://github.com/spiral-project/daybed/blob/master/daybed/views/data.py#L21, https://github.com/spiral-project/daybed/blob/master/daybed/views/definition.py#L23

leplatrem:
Note: could this be a typo ? (`header`(s)) : https://github.com/mozilla-services/cornice/blob/master/cornice/schemas.py#L92

almet:
ah right. I though I opened this to cornice. That's indeed a cornice bug.

leplatrem:
Closed in Cornice :dancers: !

PR #227, opened 2014-09-05
Reproduce CORS missing headers

leplatrem:
As a good start, ticket #48 is reproduced in a test case :)

### DIFF
--- a/daybed/views/errors.py
+++ b/daybed/views/errors.py
@@ -1,14 +1,21 @@
+from pyramid.response import Response
 from pyramid.security import Everyone
-from pyramid.httpexceptions import HTTPUnauthorized, HTTPForbidden
 
 from cornice.cors import ensure_origin
 
 
 def forbidden_view(request):
     if not request.credentials_id or request.credentials_id == Everyone:
-        response = HTTPUnauthorized()
+        resp = Response(
+            '{"error": "401 Unauthorized",'
+            ' "msg": "You must be logged-in to access this page."}',
+            status='401 Unauthorized', content_type='application/json')
     else:
-        response = HTTPForbidden()
+        resp = Response(
+            '{"error": "403 Forbidden",'
+            ' "credentials_id": "%s", "msg": "Access to this resource is '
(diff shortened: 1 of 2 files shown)


almet:
Sweet, that will help :)

Natim:
Note that we also probably want forbidden 401 to return JSON.

Natim:
According to https://github.com/spiral-project/daybed/blob/master/daybed/views/errors.py it should be the case.

Natim:
The problem is that if we provide the `Authorization` header that doesn't validates for any reason, pyramid and pyramid_hawk stops the request.

I am not sure about the strategy:
- Should we make sure daybed.js or backbone-daybed doesn't send the Authorization header for OPTIONS request's types.
- Should we allow daybed to answer OPTIONS calls even if the Authorization header is broken?

What do you guys think?

almet:
headers aren't sent in this particular case, but that's not the only problem.

I was working on a patch about that yesterday, but still a work in progress.

almet:
(the missing header is the ORIGIN one, since self.headers isn't sent as part of the request)

almet:
This latest changes should fix it.

Hope the code explains itselfs enough :)

daybed/views/errors.py
```diff
@@ -1,15 +1,21 @@
...
 
 
 def forbidden_view(request):
     if not request.credentials_id or request.credentials_id == Everyone:
-        return Response(
-            '{"error": "401 Unauthorized",'
-            ' "msg": "You must be logged-in to access this page."}',
-            status='401 Unauthorized', content_type='application/json')
-    return Response(
-        '{"error": "403 Forbidden",'
```
Natim, line 12: Does HTTPUnauthorized and HTTPForbidden returns a JSON response?

almet, line 12: I don't think so, but I don't think that should be a requirement.

the trick I did could also work if we let `Response()` and add the cornice trick just after.

almet:
updated.

leplatrem:
Great! Thanks!

PR #228, opened 2014-09-05
Fix default renderer (fixes #181)

### LINKED ISSUE #181: geojson renderer used when not needed
By default, the geojson renderer is used even when it doesn't make sense. We shouldn't do that.

leplatrem:
As a good start, ticket #181 is reproduced in a test case :)

**update** : fixed :)


Natim:
Popopop :+1:

PR #229, opened 2014-09-07
Prevent anynomous models to be stolen

leplatrem:


Natim:
If we do it like that in default_permissions, I agree because we still let the user add the update_permissions to anyone if he wants to.

leplatrem:
Great :) Merge please :)

PR #230, opened 2014-09-08
Save permissions during model creation

leplatrem:
Requires #234 
- [x] Use validated['permissions'] in views
- [ ] Default permissions if not specified
- [ ] Merge permissions if specified (always give ALL to owner)


Natim:
See #235

PR #235, opened 2014-09-12
Let POST /models and PUT /models/:model_id handle permissions configuration.

Natim:


leplatrem:
Yes! Thanks !

Maybe we can avoid duplicated code, and move it somewhere in permissions.py ?

You could add a test to check that authors permissions are always ALL, even if provided during model creation...

Natim:
> You could add a test to check that authors permissions are always ALL, even if provided during model creation...

Is it what we want though?

leplatrem:
I thought so since I could not find any usecase otherwise! But indeed we might want to keep it stupid simple.
Whatever you choose, keep consistent with PATCH/PUT permissions!

PR #240, opened 2014-10-12
Use the read the docs theme locally.

almet:


leplatrem:
Tests fail because of 502 bad gateway (travis timeout).

I merge anyway.

PR #245, opened 2014-10-14
Add a metadata field which isn't validated.

almet:


daybed/schemas/validators.py
```diff
@@ -32,7 +32,8 @@ def __init__(self, definition):
         for field in definition['fields']:
             field['root'] = self
             fieldtype = field.pop('type')
-            self.add(registry.validation(fieldtype, **field))
+            if fieldtype is not 'metadata':
+                self.add(registry.validation(fieldtype, **field))
```
leplatrem: Why was this necessary ? If a metadata has no specific validation, a colander Mapping should be quite flexible enough IMO

leplatrem:
Eventually, with @Natim example, I could get what you want to achieve !

Maybe the name `metadata` is not totally obvious... Hard to find better though I must admit :) What about `annotation`, `note`, `comment`... ?

almet:
Yeah obviously this was in need of tests. "annotation" seems to be the best fit here.

Natim:
Ok to change for annotation

Natim:
PR updated.

PR #249, opened 2014-10-27
Accept an extra property to models definition to put other model related informations.

Natim:
Fix the Formbuilder error with `submitButtonLabel`

### DIFF
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -142,7 +142,7 @@ The definition properties are:
 * **title**: The model title
 * **description**: The model description
 * **fields**: The model fields' list. See :ref:`fields documentation <fieldtypes-section>`
-* **extra**: An optional mapping to link some more information to your model.
+* **extra**: An optional property to store custom data to your model.
 
 
 **GET /models**
--- a/docs/source/usage.rst
+++ b/docs/source/usage.rst
@@ -137,6 +137,14 @@ Of course, the model permissions can be changed later.
     by a **POST**. A random name will be generated.
 
 
+The definition properties are:
+
+* **title**: The model title
+* **description**: The model description
+* **fields**: The model fields' list. See :ref:`fields documentation <fieldtypes-section>`
+* **extra**: An optional mapping to link some more information to your model.
+
+
 **GET /models**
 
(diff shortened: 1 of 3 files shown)


leplatrem:
What do you think of a single optional attribute like `extra`, instead of allowing to store anything possible ?

Natim:
Well what is the difference?

daybed/tests/test_functional.py
```diff
@@ -209,6 +209,7 @@ def valid_definition(self):
         return {
             "title": "Mushroom Spots",
             "description": "Where are they ?",
+            "submitButtonLabel": "Let's go!",
```
leplatrem: This is not the right way of testing :) This doesn't explicit the specs of what you just achieved in this commit ! Or to say it differently, if one part or another of your commit is removed, the failed test error won't be explicit at all!

It would be a lot better if you add a dedicated **unit** test here : https://github.com/spiral-project/daybed/blob/master/daybed/tests/test_schemas_validators.py#L41-L125

leplatrem:
Well, it would keep models definitions tidy.

Plus I guess that validating an object with 4096 properties is a lot more costly than validating an object with 5 :)

Natim:
I've added some documentation about the new `extra` fields.

docs/source/usage.rst
```diff
@@ -137,6 +137,14 @@ Of course, the model permissions can be changed later.
     by a **POST**. A random name will be generated.
 
 
+The definition properties are:
+
+* **title**: The model title
+* **description**: The model description
+* **fields**: The model fields' list. See :ref:`fields documentation <fieldtypes-section>`
+* **extra**: An optional mapping to link some more information to your model.
```
leplatrem: _optional property to store custom data_ may be clearer no ?

leplatrem:
Thanks for taking care of me and my narrow-minded feedbacks :)

Natim:
@leplatrem Your are more than welcome, I like it :+1:

leplatrem:
LGTM! You can merge when Travis approves ;)

almet:
thanks :)

PR #250, opened 2014-10-28
Automatically redirect to /{api version} if it's missing

magopian:
I got bitten by the documentation saying that I could go to
http://localhost:8000/fields, and had a 404.

@Natim told me that instead of fixing the code, I could automatically redirect
to use the api version if the path is missing it.

If it's a bad idea, then it's @Natim's fault ;)

(Obviously, if it's a good idea, it's thanks to me, right? ;)

### DIFF
--- a/daybed/__init__.py
+++ b/daybed/__init__.py
@@ -12,6 +12,7 @@ logger = logging.getLogger(__name__)
 
 import six
 from cornice import Service
+from pyramid import httpexceptions
 from pyramid.config import Configurator
 from pyramid.events import NewRequest
 from pyramid.renderers import JSONP
@@ -26,6 +27,9 @@ from daybed.permissions import (
 from daybed.views.errors import forbidden_view
 from daybed.renderers import GeoJSON
 from daybed import indexer, events
+
+
+API_VERSION = 'v%s' % __version__.split('.')[0]
 
 
 def settings_expandvars(settings):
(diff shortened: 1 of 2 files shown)


Natim:
Why did you amend the master commit?

Natim:
I think the documentation is another question, but at least this will be useful even if we want to fix the documentation later.

Natim:
I have no specific comments to add to this PR, it looks good to me.

leplatrem:
Yes thanks ! Welcome onboard @magopian !

More generally I'm suprised that pyramid doesn't provide url rewrite (see hard coded urls in the code: https://github.com/spiral-project/daybed/blob/master/daybed/views/models.py#L115-L116)

magopian:
aouch, indeed :/ and thanks ;)

daybed/__init__.py
```diff
@@ -45,6 +49,12 @@ def build_list(variable):
     return [v.strip() for v in variable]
 
 
+def redirect_to_version(request):
+    """Redirect to the current version of the API."""
+    raise httpexceptions.HTTPTemporaryRedirect(
```
almet, line 54: Thanks for using a 307 here :)

magopian, line 54: Yeah, it was my idea (of course it wasn't... @natim told me that you would want that, so I did ;)

PR #256, opened 2014-11-03
Store deserialized version of nodes

leplatrem:


daybed/schemas/object.py
```diff
@@ -55,8 +54,7 @@ def definition(cls, **kwargs):
     @classmethod
     def validation(cls, **kwargs):
         definition = cls._fetch_definition(kwargs)
-        kwargs['validator'] = RecordValidator(definition)
-        return super(ObjectField, cls).validation(**kwargs)
+        return super(ObjectField, cls).validation(definition, **kwargs)
```
leplatrem, line 65: This particular line is not very PEP20 :( 

Using this https://github.com/spiral-project/daybed/pull/256/files#diff-a1faa1fde6616ef6569165d485c880f3R109, it will instantiate a `RecordSchema(definition)`. A comment might be enough, or do you have a better idea ?

daybed/schemas/list.py
```diff
@@ -1,24 +1,14 @@
...
-    def __call__(self, node, value):
-        self.schema.deserialize(value)
+from .json import JSONSequence
 
 
 @registry.add('list')
 class ListField(TypeField):
-    node = JSONList
-    hint = _('A list of objects')
+    hint = _('A list of values')
```
Natim, line 11: `values or objects.`

daybed/schemas/object.py
Natim, line 65: `RecordSchema.validation(cls, definition, **kwargs)` would not work?

Natim, line 65: Why did you choose to refactor this?

daybed/schemas/list.py
leplatrem, line 11: values.

which can be int, string, ... or objects !

daybed/tests/test_schemas_object.py
```diff
@@ -128,13 +140,13 @@ def test_is_defined_with_an_existing_model(self):
         self.assertDictEqual(self.definition, field)
 
     def test_validation_succeeds_if_fields_are_valid(self):
-        value = self.validator.deserialize('{"age": 12}')
```
Natim, line 131: I think we should test both because `'{"age": 12}'` is valid JSON

daybed/schemas/object.py
leplatrem, line 65: Nope, we need the `kwargs` transformation of `TypeField.validation()`

daybed/tests/test_schemas_object.py
```diff
@@ -128,13 +140,13 @@ def test_is_defined_with_an_existing_model(self):
...
+        value = self.validator.deserialize({"age": "12"})
         self.assertDictEqual(value, {'age': 12})
 
     def test_validation_fails_if_fields_is_invalid(self):
         self.assertRaises(colander.Invalid,
                           self.validator.deserialize,
-                          '{"age": "a"}')
```
Natim, line 137: Are you not validating JSON?

daybed/tests/test_schemas_object.py
```diff
@@ -128,13 +140,13 @@ def test_is_defined_with_an_existing_model(self):
...
+                          {"age": "a"})
```
Natim: This is not JSON anymore it is Python dict.

daybed/schemas/list.py
Natim, line 11: Ok

daybed/schemas/object.py
Natim, line 65: Well I suppose a comment would make it then

daybed/tests/test_schemas_object.py
leplatrem: The object field does not have to be JSON actually.

leplatrem:
Ready to be reviewed/merged :)

PR #257, opened 2014-11-04
Create/update model via definition endpoint (fixes #231)

### LINKED ISSUE #231: Add definition PUT, that would create/update models
Would allow JS libs to fetch/save on the same URL

leplatrem:
Allow to GET/PUT definitions on the same API end point

### DIFF
--- a/daybed/schemas/relations.py
+++ b/daybed/schemas/relations.py
@@ -14,7 +14,7 @@ class ModelExist(object):
 
     def __call__(self, node, value):
         try:
-            self.db.get_model_definition(value)
+            self.db.get_model(value)
         except ModelNotFound:
             msg = u"Model '%s' not found." % value
             raise Invalid(node, msg)
--- a/daybed/schemas/relations.py
+++ b/daybed/schemas/relations.py
@@ -14,7 +14,7 @@ class ModelExist(object):
 
     def __call__(self, node, value):
         try:
-            self.db.get_model(value)
+            self.db.get_model_definition(value)
         except ModelNotFound:
             msg = u"Model '%s' not found." % value
             raise Invalid(node, msg)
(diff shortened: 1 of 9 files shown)


daybed/schemas/relations.py
```diff
@@ -14,7 +14,7 @@ def __init__(self, db):
 
     def __call__(self, node, value):
         try:
-            self.db.get_model_definition(value)
+            self.db.get_model(value)
```
Natim: I am not understanding why you need to change that and create this method to the backend.

daybed/views/models.py
```diff
@@ -44,6 +46,25 @@ def get_definition(request):
...
+    """Create or update a model definition."""
+    model_id = request.matchdict['model_id']
+    try:
+        model = request.db.get_model(model_id)
+    except ModelNotFound:
+        model = {}
+
+    permissions = invert_permissions_matrix(model.get('permissions', {}))
+    model['permissions'] = permissions
+    model['records'] = []
```
Natim: How does this works with this: https://github.com/spiral-project/daybed/pull/257/files#diff-feded148953d5e4784370906521a1bd2R217 ?

daybed/schemas/relations.py
leplatrem: It was just to retrieve definition and permission in one call. Not entirely necessary indeed... I can get rid of it

daybed/views/models.py
leplatrem: the `handle_put_model` method will just iterate the empty records collection, but won't erase the existing records

Natim: Ok

leplatrem:
Thanks @Natim, the diff is much better now :)

Natim:
;)

PR #260, opened 2014-11-05
Add an id generator which generates memorable words.

almet:


setup.py
```diff
@@ -36,28 +36,30 @@
...
 REQUIREMENTS = [
-    'setuptools',
-    'six',
-    'redis',
+    'CouchDB==0.10.1dev',
+    'colander',
+    'cornice',
+    'elasticsearch',
     'hiredis',
+    'koremutake==1.1.0',
```
leplatrem: Are we obliged to keep version numbers here ?

daybed/tests/test_id_generators.py
```diff
@@ -0,0 +1,33 @@
...
+        encode.side_effect = ['existing-value', 'new-value']
+        created = ['existing-value']
+        def _exists(key):
+            return key in created
+
+        generator = KoremutakeGenerator()
+        self.assertEquals(generator(key_exist=_exists), 'new-value')
+
+    def test_it_returns_a_string_with_a_max_size(self):
+        generator = KoremutakeGenerator()
```
leplatrem: since you instantiate the same generator in each test method, put it in setUp()

setup.py
almet: I believe so, but haven't tried without it. Note that this isn't the released version but a version that relies on my fork.

daybed/backends/id_generators.py
```diff
@@ -8,5 +11,25 @@ class UUID4Generator(object):
...
+    def __init__(self, config=None):
+        if config is None:
+            settings = {}
+        else:
+            settings = config.registry.settings
+
+        self.max_bytes = int(settings.get('id_generator.max_bytes', 4))
+
+    def __call__(self, key_exist=None, tries=5):
+        random_int = int(os.urandom(self.max_bytes).encode('hex'), 16)
```
almet: It seems this doesn't work with py3.4. Any idea on how to have something that works for both py3k and py2 ?

leplatrem:
You didn't update memorybackend, was it on purpose ?

conf/development.ini
```diff
@@ -8,7 +8,9 @@ daybed.backend = daybed.backends.redis.RedisBackend
 elasticsearch.hosts = localhost:9200
 
 # Model name generator configuration
-daybed.id_generator = daybed.backends.id_generators.UUID4Generator
+daybed.id_generator = daybed.backends.id_generators.KoremutakeGenerator
+id_generator.max_bytes = 4
```
leplatrem, line 12: Will we want to ever configure this ?

conf/development.ini
```diff
@@ -8,7 +8,9 @@ daybed.backend = daybed.backends.redis.RedisBackend
 elasticsearch.hosts = localhost:9200
 
 # Model name generator configuration
-daybed.id_generator = daybed.backends.id_generators.UUID4Generator
+daybed.id_generator = daybed.backends.id_generators.KoremutakeGenerator
```
leplatrem, line 11: you can set it in docker too

daybed/tests/test_id_generators.py
almet: That's correct. Updated.

conf/development.ini
almet, line 12: I believe so, at some point, in case we have too many collisions (because we have too many users)

setup.py
leplatrem: Yes, I saw it in dependency links. By experience, pinning versions in setup files can lead to headache :)

almet: I can remove it and test if that works :)

almet:
Not it's not, I just forgot that!

almet:
Updated again. Still misses the py3k fix.

leplatrem:
For py3k, you could catch `AttributeError`, see https://github.com/elasticsearch/elasticsearch-py/commit/18d6b41e8299d3e62bfac2a8c2f8ceb87be24f45

PR #261, opened 2014-11-07
Add Basic Auth to POST /tokens to let the user get the same token for the same user:password

Natim:


almet:
Yeah! We've got our login! oh yeah. Usability++

leplatrem:
Nice :)

almet:
I think we should expose this very easily into daybed.js.

daybed/views/tokens.py
```diff
@@ -10,10 +14,26 @@
 @tokens.post(permission='post_token')
 def post_tokens(request):
     """Creates a new token and store it"""
-    token, credentials = get_hawk_credentials()
-    request.db.store_credentials(token, credentials)
 
-    request.response.status = "201 Created"
+    # If we have an authorization header with the Basic or Token realm
+    # Use it as the base HKDF for building the same token
```
almet: "use it to derive the key"

daybed/views/tokens.py
```diff
@@ -10,10 +14,26 @@
...
+    session_token = None
+    if request.authorization and \
+       request.authorization[0] in ["Basic", "Token"]:
+        session_token = hmac.new(
+            request.registry.tokenHmacKey.encode("ascii"),
+            ("%s %s" % request.authorization[:2]).encode("utf-8"),
+            hashlib.sha256
+        ).hexdigest()
```
leplatrem: Shouldn't we move this to the `daybed.tokens` module instead ? (imports are a good hint :))

Natim: Yes it is probably a good idea.

PR #266, opened 2014-11-11
Search end point improvements

leplatrem:
Required for https://github.com/spiral-project/daybed.js/pull/28 to work 
- [x] Allow POST on search, because couldn't figure out how to set a body on a GET request with XMLHttpRequest
- [x] Prevent catching ES query syntax errors

### DIFF
--- a/daybed/tests/test_views.py
+++ b/daybed/tests/test_views.py
@@ -656,6 +656,9 @@ class SearchViewTest(BaseWebTest):
                           headers=self.headers)
 
     def test_search_returns_200_if_query_is_correct(self):
+        self.app.post('/models/test/search/', {'match_all': {}},
+                      headers=self.headers,
+                      status=200)
         self.app.get('/models/test/search/', {'match_all': {}},
                      headers=self.headers,
                      status=200)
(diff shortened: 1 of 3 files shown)


daybed/tests/test_views.py
```diff
@@ -692,6 +695,16 @@ def test_search_returns_502_if_elasticsearch_fails(self, search_mock):
                      headers=self.headers,
                      status=502)
 
+    @mock.patch('elasticsearch.client.Elasticsearch.search')
+    def test_search_returns_original_code_on_bad_request(self, search_mock):
+        import elasticsearch
```
almet: Why importing elasticsearch here rather than at top level?

daybed/views/search.py
```diff
@@ -33,6 +35,9 @@ def search_records(request):
     try:
         results = request.index.search(model_id, query, params=params)
         return results
+    except elasticsearch.TransportError as e:
```
almet: I would expect this to be done inside the `search` method of the index rather than here, to keep the view file independent from elasticsearch.

Otherwise, we can define a SearchError exception that is to be used by the indexes?

leplatrem: +1

almet:
That looks good to me!

PR #267, opened 2014-11-12
Improve the ElasticSearch cluster configuration to allow basic_auth ES access

Natim:


daybed/indexer.py
```diff
@@ -16,7 +17,46 @@ def __init__(self, *args):
...
+                if arguments.scheme == 'https':
+                    use_ssl = True
+                    port = 443
+
+            self.built_hosts.append({
+                'host': hostname,
+                'port': int(port),
+                'use_ssl': use_ssl,
+                'http_auth': http_auth
+            })
```
leplatrem: Isn't there any python lib that could handle this for us?
If not, I suggest you could at least move this piece of code to some utility function

daybed/indexer.py
```diff
@@ -16,7 +17,46 @@ def __init__(self, *args):
 class ElasticSearchIndexer(object):
 
     def __init__(self, hosts, prefix):
-        self.client = elasticsearch.Elasticsearch(hosts)
+        self.built_hosts = []
```
leplatrem: why do you store it as attribute ?

Natim: It was to test it but I suppose I can mock the Elasticsearch constructor?

Natim:
Updated.

Makefile
```diff
@@ -41,3 +41,11 @@ tests-failfast: install-dev
...
 serve: install install-dev
 	$(VENV)/bin/pserve conf/development.ini --reload
+
+.PHONY: spaceleft
+spaceleft:
+	@if which grin 2>&1 >/dev/null; \
+	then \
+      test "$$(grin " $$" daybed/ docs/ -l | wc -l)" -ne "0"  && \
+	  grin -l " $$" daybed/ docs/ | xargs sed -i 's/\s*$$//' || exit 0; \
+	fi
```
leplatrem: I don't really see the point of this here, you could define it in your shell instead, no ?
(.phony is repeated, is that ok ?)

leplatrem:
I'm not against merging this, but a bit doubtful though.

We can have all this code in Daybed, but in the mean time, opening a PR on elasticsearch-py would be worth it IHMO. I don't see why it shouldn't support absolute urls for hosts!

Natim:
Why should it?

leplatrem:
The dict formalism is rather curious, since there's a standard for that! RFC-1738 :)

Natim:
I can make a pull-request to them based on this assumption. There is not urgency yet to merge this patch since I already applied it in production.

leplatrem:
Ok.

Also, I wouldn't have created a dedicated module just for one utility function. But it may be a matter of taste :) 
More generally, we should start thinking of #213 :)

Natim:
> More generally, we should start thinking of #213 :)

Exactly the reason why I did it like so :)

Natim:
Here we go: https://github.com/elasticsearch/elasticsearch-py/pull/149

leplatrem:
Good job @Natim! Your PR was merged in ES python client :) We will soon be able to close this here!

PR #114, opened 2014-11-21
We shouldn't need the same permissions for a PUT which create or modify a record same for models.

Natim:


Natim:
I had this bug with https://github.com/spiral-project/daybed-cloud-share, I am going to fix it.

leplatrem:
I'm not sure about what is done in this PR actually :(

Could explicit your specifications in the tests please ? 

I would have expected something like :
- test that a user without creation permission cannot create a record with PUT
- test that a user without update permission can create a record with PUT

And since the title says "same for models", I would have expected the same for models :)

Please note that views permissions tests are not supposed to be functional test suites, something in RecordsViewsTest would be more efficient, since not executed several times with different backend etc.

Natim:
Ok I will rework this tests.

PR #271, opened 2014-11-26
Add a blob field that is not being indexed.

almet:


leplatrem:
Shouldn't we add limit of size btw ?

Also, would you mind adding a small mention in docs ?

Natim:
IMHO, I don't think the size should be limited on the field but maybe we will add some other kind of limitation in the server configuration for that.

